PR #2, opened 2020-04-10
fix git clone github url

hellkite500:
## Changes

- Point git clone url to NOAA-OWP instead of fork

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -55,7 +55,7 @@ The following sequence of commands should provide a sense of the operation of th
 
 ```
 pip3 install geopandas numpy pandas xarray netcdf4 
-git clone --progress --single-branch --branch master http://github.com/jameshalgren/t-route.git
+git clone --progress --single-branch --branch master http://github.com/NOAA-OWP/t-route.git
 cd src/fortran_routing/mc_pylink_v00/MC_singleCH_singleTS/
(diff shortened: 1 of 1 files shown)


jameshalgren:
@Nels Frazier - NOAA Affiliate <nels.frazier@noaa.gov>
Squash and merge or rebase for a small change like this?
(your prior comment seems to indicate "always rebase"...)
--
James Halgren
NOAA National Water Center
+1 205-347-1453
Lynker <https://www.lynker.com/>

hellkite500:
In this case they are one and the same.  Rebase just keeps all commits in the PR and applies them to the head of the destination branch.  Since this is only one commit, it will do the same.  Only difference is squash and merge will generate a new commit id and a new commit message to associate with it.  If there are multiple commits, squashing creates a new single commit with all changes.  Sometimes this is required to prevent merge conflicts, though typically you do this locally and update the PR with whatever commit you want in the upstream history.  If you don't want certain commits to show up in the upstream history, they should be rebased in the fork/local dev clone then the "final" ones pushed to the PR branch.

jameshalgren:
Thanks -- that make sense. Since it's the same, I squashed for this time.
--
James Halgren
NOAA National Water Center
+1 205-347-1453
Lynker <https://www.lynker.com/>

Issue #5, opened 2020-04-17
Include reservoirs in the parsing of connections


jameshalgren:
The current algorithm carries the network across reservoirs. 

The code in `./src/python_framework` needs to be updated to identify the routing breaks that occur at reservoirs.

jhrehanoaa:
closed by #38

PR #18, opened 2020-04-27
Add waterbodies to framework base functions

jameshalgren:
Water bodies are now parsed and returned in the 'supernetwork values' collection returned by the nhd_network_utilities function. These are not used further by this version of code but a later version will use the waterbodies to add further discretization to the network traversal.


jameshalgren:
Closes #5

jameshalgren:
Juzer,
Is it correct that the change to WPC affects only the qdc and not the velc
and depthc?
--
James Halgren
NOAA National Water Center
+1 205-347-1453
Lynker <https://www.lynker.com/>

src/python_framework/networkbuilder.py
```diff
@@ -33,9 +32,58 @@ def get_down_connections(
...
+    , data_key = r'data'
+    , downstream_key = r'downstream'
+    , upstreams_key = r'upstreams'
+    , verbose = False
+    , debuglevel = 0
+    ):
+
+    if verbose: print('waterbody_set ...')
+    waterbody_set = {con[data_key][waterbody_col] for key, con in connections.items()}
+    waterbody_set.remove(waterbody_null_code)
```
groutr, line 49: this should probably use `set.discard`, unless you want an error thrown if the null code doesn't exist.

jameshalgren:
@jdhondia-noaa I just recommitted with the full correction to WPC using proper declaration and passing in the fortran code so that WPC is not used uninitialized.

src/python_framework/networkbuilder.py
jameshalgren: Thanks -- made that change.

Issue #11, opened 2020-04-21
Optimize data storage for better performance


jameshalgren:
Received this comment: 
>I think the structure of the connections dictionary needs to be reconsidered. I switched a few parts of the implementation to be array based and use slicing instead of loops.
My latest experiments have a multiprocessed, 1440 timestep, run executing in just under a minute on my laptop. The serial version runs in about 1m45s.

jameshalgren:
@groutr Thanks for this issue. When it works we'd love a pull request.

groutr:
This was experimented with using the single reach code.
The main ideas here are:
1. Separate the graph from the data. This simplified a many of the graph operations
2. Keep the data as a single numpy array, sorted by node id. Keeping things together in memory helped a lot IIRC. Using numpy types rather than python objects also reduced memory consumption quite a bit. The sorting was important because we used a binary search to generate masks for subsets of the graph.
3. Use a numpy array to track flowdepth. The index corresponded to the index of the sorted data array.

By using numpy arrays, slicing could be used to eliminate almost all of the looping in the single reach code.

jameshalgren:
Closed with #110 (see issue #11)

PR #41, opened 2020-05-26
New network representation and array based computations

groutr:
This is a work-in-progress. Do not merge into master.

This PR experiments with a new network implementation, a continuation of #11. This PR adds two new modules to the framework, `nhd_network` and `nhd_io`. `nhd_network` has general utilities for manipulating an adjacency list as well as a topological sort using Kahn's algorithm.
We use a topological sort to determine the order in which to process edges in the graph instead of decomposing the graph into reaches.

The correctness of this code still needs to be verified.


jameshalgren:
Per "do not merge" comment, converted this to a draft PR.

Issue #14, opened 2020-04-22
convert to command-line inputs


jameshalgren:
Many of the test scripts (e.g. nhd_network_traversal.py, nhd_reach_parallel_test.py) have debug_level, verbose, and supernetwork selection. Make these command-line inputs so that commits do not have to track any testing changes.

jameshalgren:
PR #29 is working on this. There will be a few additional PRs to finish this issue.

jameshalgren:
Need to use the argparse feature `choices` (see fifth program block here: https://docs.python.org/2/howto/argparse.html#combining-positional-and-optional-arguments)

jhrehanoaa:
Should we consider this issue technically closed for now?

jhrehanoaa:
Closed by #42

PR #42, opened 2020-05-26
Argparse

jhrehanoaa:
Added argparse to nhd_reach_parallel_test and nhd_network_traversal. Updated utilities file to ingest custom input file. Various cleanup to command-line arguments. Closes#27


src/python_framework/nhd_reach_parallel_test.py
```diff
@@ -12,7 +12,8 @@ def _handle_args():
         "--debuglevel",
         help="Set the debuglevel",
         dest="debuglevel",
-        choices=["0", "-1", "-2", "-3"],
+        type = int,
+        choices=["0", "1", "2", "3"],
```
jameshalgren, line 16: thanks!

jameshalgren:
> Let's load this diff, then I think we will be ready:
> 
> [argparse_suggest.txt](https://github.com/NOAA-OWP/t-route/files/4691465/argparse_suggest.txt)

Executed the following tests: 
1. Using the built-in networks
```
python nhd_network_traversal.py -v -n Brazos_LowerColorado_Named_Stream
```
1. Using a custom network (the sample file, in this case...)
```
python nhd_network_traversal.py -v -n custom -f ../../test/input/json/CustomInput.json
```
1. Using the `custom` flag without a path (Returns error -- path to .json file required if 'custom' supernetwork selected.)
```
python nhd_network_traversal.py -v -n custom 
```

jameshalgren:
> Let's load this diff, then I think we will be ready:
> 
> [argparse_suggest.txt](https://github.com/NOAA-OWP/t-route/files/4691465/argparse_suggest.txt)

@jhrehanoaa 
We should also apply that to the other files with arg parsing (and I will open an issue about consolidating that into a single source file.)

jameshalgren:
closes #14

PR #47, opened 2020-05-28
Fix initialization of variables.

groutr:
Attempts to fix the initialization of input parameters. This PR results in repeatable results for me.
~~For the moment, `assume_short_ts` is being disabled.~~

@jameshalgren, would you please look this over and verify that things are initialized to the proper values?

### DIFF
--- a/src/python_routing_v02/compute_nhd_routing_SingleSeg.py
+++ b/src/python_routing_v02/compute_nhd_routing_SingleSeg.py
@@ -283,6 +283,9 @@ def compute_mc_reach_up2down(
         current_flow["vel"]["prev"] = current_flow["vel"]["curr"]
         current_flow["qlat"]["prev"] = current_flow["qlat"]["curr"]
 
+        if assume_short_ts:
+            quc = qup
+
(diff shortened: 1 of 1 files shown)


src/python_routing_v02/compute_nhd_routing_SingleSeg.py
```diff
@@ -213,21 +213,15 @@ def compute_mc_reach_up2down(
...
                 writeString = writeString + f"\n upstream seg : {us}"
-            qup_tmp += flowdepthvel[us]["flow"]["curr"]
-            if assume_short_ts:
-                quc_tmp = qup_tmp
-            else:
-                quc_tmp += flowdepthvel[us]["flow"]["curr"]
+            qup += flowdepthvel[us]["flow"]["prev"]
+            quc += flowdepthvel[us]['flow']['curr']
     if write_output:
         writetoFile(file, writeString)
```
jameshalgren, line 227: breaks with '-o' flag because of qup_tmp in write statement on line 234 in current version.

Issue #43, opened 2020-05-27
Consolidate argparse into a single script


jameshalgren:
The recently submitted argparse changes could be consolidated into a single argument handling script.

jameshalgren:
Some tips on custom 'usage' instructions are in this S.O. post: 
https://stackoverflow.com/questions/15935092/creating-mutually-inclusive-positional-arguments-with-argparse

jhrehanoaa:
I moved most of the code over to a single script. The nhd_reach_parallel file is getting hung up on the reservoir inputs though.

jhrehanoaa:
closed by #55

PR #55, opened 2020-06-10
New argparse

jhrehanoaa:
This contains the separate command line file nwm_commandline that allows custom network inputs with json locations. Additionally, the break at reservoirs feature is contained within the code as well. Ignore commit message that reservoir feature does not work.  

Closing #43


notebooks/LR_NetCDF.ipynb
```diff
@@ -0,0 +1,680 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 17,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "\n",
+    "import xarray as xr\n",
+    "\n",
+    "ds = xr.open_dataset('/home/APD/inland_hydraulics/wrf-hydro-run/OUTPUTS/202003191900.CHRTOUT_DOMAIN1')\n",
```
jameshalgren, line 12: We need to keep machine specific paths like this one out of commits

notebooks/LR_NetCDF.ipynb
```diff
@@ -0,0 +1,680 @@
...
+    "df = ds.to_dataframe()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 18,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
```
jameshalgren, line 22: Let's save the file without outputs for version control.

src/python_framework/nhd_reach_parallel_test.py
```diff
@@ -2,72 +2,9 @@
 import sys
 import os
 import time
-
+import nwm_network_commandline as cmd
+import argparse
```
jameshalgren, line 6: I don't think we'll need the argparse import in this file now.

src/python_framework/nwm_network_commandline.py
```diff
@@ -0,0 +1,84 @@
...
+        choices=["0", "1", "2", "3"],
+        default=0,
+    )
+    parser.add_argument(
+        "-w",
+        "--break_at_waterbodies",
+        help="Use the waterbodies in the route-link dataset to divide the computation (leave blank for no splitting)",
+        dest="break_network_at_waterbodies",
+        action="store_true",
+    )
```
jameshalgren, line 24: This is probably something we should eventually talk to Nels about and see if we can use this parser with some customization. (Some flags are not relevant, depending on the file). But this will work for now, with the noted comments fixed.

notebooks/LR_NetCDF.ipynb
jameshalgren: If possible, this could be separated into two different files.

PR #57, opened 2020-06-12
Make Fortran Precision Explicit

jameshalgren:
The fortran source files providing the access to the muskingum cunge routing engine were all ambiguously typed and calculation precision was subject to the defaults of the compiler. F2PY, which we have used for our efforts to this point is also somewhat ambiguous -- the documentation seems to indicate that it can handle a declared double precision variable, but in testing, we could not get a clear replication of the behavior in such a way that we had confidence in the use of double precision floating point numbers passing through that package. A couple of the links below provide information that can help provide additional (though hardly exhaustive) context.

To test, use the mc_sseg_stime_NOLOOP.py and mc_srch_stime.py files in the fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS folder.


jameshalgren:
Links:
https://stackoverflow.com/questions/10935126/subroutine-argument-not-passed-correctly-from-python-to-fortran
https://stackoverflow.com/questions/12523524/f2py-specifying-real-precision-in-fortran-when-interfacing-with-python
http://fortranwiki.org/fortran/show/Real+precision
https://scicomp.stackexchange.com/questions/910/how-to-set-double-precision-values-in-fortran
https://stackoverflow.com/questions/6146005/is-there-a-better-double-precision-assignment-in-fortran-90

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/mc_sseg_stime_NOLOOP_demo.py
```diff
@@ -90,33 +102,73 @@ def main():
...
+        0 1 0.06863606721162796 0.07104580104351044 0.009998836554586887
+        0 2 0.04598825052380562 0.06900732219219208 0.009981763549149036
+        0 3 0.21487340331077576 0.07048019021749496 0.010033470578491688
+        1 0 0.3173820674419403 0.15248453617095947 0.03275299444794655
+        1 1 0.2014089822769165 0.09322302788496017 0.015030190348625183
+        1 2 0.13257014751434326 0.061184611171483994 0.008333344012498856
+        1 3 0.7570106983184814 0.12373604625463486 0.02334451675415039
+        """
+        # Providing the truncated inputs produces the exact same answer...
+        qup = 0.04598825
```
groutr, line 122: I think this should use the ctypes.c_float type. qup is still a double precision float here (even though it is truncated).
```
import ctypes
qup = ctypes.c_float(0.04598825052380562)
```

groutr: Hmm, f2py doesn't seem to like ctypes objects.
```
ValueError: mc_sseg_stime.muskingcunge_module.muskingcungenwm() 1st argument (dt) can't be converted to float
```

jameshalgren: Based on your experiment here, I think we're moving forward by forcing the precision in the fortran to single precision. We can maybe take a second step to make sure that the communication between the modules is explicitly single or double precision.

PR #65, opened 2020-06-24
Script to fixup noteboks prior to committing.

groutr:
This script will:
  * Clear output cells
  * Clear Python metadata
  * Try to reformat code blocks with black

ping: @jameshalgren

### DIFF
--- a/notebooks/formatnb.py
+++ b/notebooks/formatnb.py
@@ -12,7 +12,7 @@ def clean_outputs(nb):
 
 def black_code(nb):
     fm = black.FileMode()
-    for cellnum, cell in enumerate(nb.cells):
+    for cellnum, cell in enumerate(nb.cells, 1):
         if cell["cell_type"] == "code":
             try:
                 cell["source"] = black.format_str(cell["source"], mode=fm)
--- a/notebooks/formatnb.py
+++ b/notebooks/formatnb.py
@@ -12,11 +12,13 @@ def clean_outputs(nb):
 
 def black_code(nb):
     fm = black.FileMode()
-    for cell in nb.cells:
+    for cellnum, cell in enumerate(nb.cells):
         if cell["cell_type"] == "code":
             try:
                 cell["source"] = black.format_str(cell["source"], mode=fm)
             except black.InvalidInput:
+                print(f"Possible invalid python encountered in cell {cellnum}")
+                print(f"Cell source causing exception:\n{cell['source']}\n")
                 continue
     return nb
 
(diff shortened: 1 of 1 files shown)


notebooks/formatnb.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def black_code(nb):
+    fm = black.FileMode()
+    for cell in nb.cells:
+        if cell["cell_type"] == "code":
+            try:
+                cell["source"] = black.format_str(cell["source"], mode=fm)
+            except black.InvalidInput:
+                continue
```
groutr, line 20: Perhaps print something about possibly invalid python code encountered?

jameshalgren: The %cell_magic lines and !shell_execution lines trigger that exception...

jameshalgren: Something like this: 
```
+++ b/notebooks/formatnb.py
@@ -1,5 +1,6 @@
 import nbformat
 import black
+from sys import exit


 def clean_outputs(nb):
@@ -12,12 +13,19 @@ def clean_outputs(nb):

 def black_code(nb):
     fm = black.FileMode()
-    for cell in nb.cells:
+    for cellnum, cell in enumerate(nb.cells):
         if cell["cell_type"] == "code":
             try:
                 cell["source"] = black.format_str(cell["source"], mode=fm)
             except black.InvalidInput:
-                continue
+                print(f"Possible invalid python encountered in cell {cellnum}")
+                print(f"cell source causing exception:\n{cell['source']}\n")
+                print(r"Script will abort (leaving your file unchanged) unless you type 'continue'")
+                abort = input()
+                if abort == 'continue':
+                    continue
+                else:
+                    exit()
     return nb
```

groutr: Notebook magics are not valid Python. Most cell magics can be isolated to their own cell.

I also don't think we want to make the tool interactive. It shouldn't add cognitive overhead to the user.

notebooks/formatnb.py
```diff
@@ -12,11 +12,13 @@ def clean_outputs(nb):
 
 def black_code(nb):
     fm = black.FileMode()
-    for cell in nb.cells:
+    for cellnum, cell in enumerate(nb.cells):
         if cell["cell_type"] == "code":
             try:
                 cell["source"] = black.format_str(cell["source"], mode=fm)
             except black.InvalidInput:
+                print(f"Possible invalid python encountered in cell {cellnum}")
```
groutr, line 20: You'll probably want `enumeate(nb.cells, 1)` to start the numbering at 1.

PR #76, opened 2020-07-09
[WIP - Confirm/Close] Add waterbody splitting

groutr:
Add waterbody splitting/handling capability to t-route.

ping @jameshalgren


jameshalgren:
*Note* 
The splitting of the modules now requires more complex compilation -- need to update `compile_singleSeg_cython.md` and probably move it to the python_routing_v02
We need to support, at a minimum the pocono2 test, which reads the test qlateral and should not split on waterbodies (or can, but there aren't any, so we can make sure that it doesn't crash for that case...)
The next step would be to read the pocono1 test, which reads the test WRF-Hydro outputs from the 
Finally, we would want to trigger the runs with a .json input file.

jameshalgren:
#115 and #116 are picking up portions of this PR... stand by.

src/python_framework_v02/nhd_network.py
```diff
@@ -264,3 +268,57 @@ def kahn_toposort_edges(N):
...
+            # find outflows of waterbody
+            outflows = set()
+            for tw in nhd_network.tailwaters(waterbody_nets[wbody_code]):
+                if tw in connections:
+                    outflows.update(connections[tw])
+            new_conn[wbody_code] = list(outflows)
+        else:
+            # copy to new network unchanged
+            new_conn[n] = connections[n]
+    return new_conn
```
jameshalgren: This block (lines 298-324) appears to be already in the main repository here:  
https://github.com/NOAA-OWP/t-route/blob/63e1578af5edf7086df7a4106314d5f31bfdbc71/src/python_framework_v02/troute/nhd_network.py#L404-L428

src/python_framework_v02/nhd_network.py
```diff
@@ -307,15 +311,15 @@ def replace_waterbodies_connections(connections, waterbodies):
...
+    #reversed_conns = reverse_network(connections)
     for n in connections:
         if n in waterbodies:
             wbody_code = waterbodies[n]
             if wbody_code in new_conn:
                 continue
 
             # get all nodes from waterbody
-            wbody_nodes = [k for k, v in waterbodies.items() if v == wbody_code]
+            wbody_nodes = nodes(waterbody_nets[wbody_code])
```
jameshalgren: @groutr I see we made this change here but the master reflects the original. Should we bring this over?

See [here...](https://github.com/NOAA-OWP/t-route/blob/63e1578af5edf7086df7a4106314d5f31bfdbc71/src/python_framework_v02/troute/nhd_network.py#L411)

jameshalgren:
Picking through these commits, to confirm whether the relevant elements of this PR have been incorporated ...

src/python_framework_v02/nhd_network.py
```diff
@@ -192,18 +192,22 @@ def dfs_decomposition(N, path_func, source_nodes=None):
...
     for h in source_nodes:
-        stack = [(h, iter(N[h]))]
+        stack = [(h, N[h])]
         while stack:
             node, children = stack[-1]
-            try:
-                child = next(children)
+            ci = child_index[node]
+            if ci < len(children):
+                child = children[ci]
```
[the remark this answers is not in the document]
groutr: No. this was likely a test to see if using lists vs iterators was faster. Using the underlying lists provides no measurable performance benefit.

src/python_framework_v02/nhd_network.py
groutr: Maybe. I think this just narrows down the number of nodes.

jameshalgren:
I think we can close this PR.

PR #77, opened 2020-07-10
[WIP] Integrating netcdf write

jdhondia-noaa:
netcdf write capabilities added to framework routing (v01)

### DIFF
--- a/src/python_routing_v01/compute_nhd_routing_SingleSeg.py
+++ b/src/python_routing_v01/compute_nhd_routing_SingleSeg.py
@@ -264,7 +264,6 @@ def compute_mc_reach_up2down(
         tw = data[supernetwork_data["topwidth_col"]]
         twcc = data[supernetwork_data["topwidthcc_col"]]
         dx = data[supernetwork_data["length_col"]]
-        bw = data[supernetwork_data["bottomwidth_col"]]
         n_manning = data[supernetwork_data["manningn_col"]]
         n_manning_cc = data[supernetwork_data["manningncc_col"]]
         cs = data[supernetwork_data["ChSlp_col"]]
--- a/src/python_routing_v01/compute_nhd_routing_SingleSeg.py
+++ b/src/python_routing_v01/compute_nhd_routing_SingleSeg.py
@@ -15,9 +15,11 @@ import multiprocessing
 import os
 import sys
 import time
+import csv
+import netCDF4
 import numpy as np
 import argparse
-
+from datetime import datetime
 
 def _handle_args():
     parser = argparse.ArgumentParser(
@@ -132,6 +134,7 @@ else:
 connections = None
 networks = None
 flowdepthvel = None
+#WRITE_OUTPUT = False  # True
 
 ## network and reach utilities
 import nhd_network_utilities_v01 as nnu
@@ -139,8 +142,10 @@ import nhd_reach_utilities as nru
 
 
 def writetoFile(file, writeString):
-    file.write(writeString)
-    file.write("\n")
+    file.write(writeString + '\n')
+    file.flush()
+    os.fsync(file.fileno())
+
 
 
 def compute_network(
@@ -155,9 +160,6 @@ def compute_network(
     global connections
     global flowdepthvel
 
-    # = {connection:{'flow':{'prev':-999, 'curr':-999}
-    #                            , 'depth':{'prev':-999, 'curr':-999}
-    #                            , 'vel':{'prev':-999, 'curr':-999}} for connection in connections}
 
     # print(tuple(([x for x in network.keys()][i], [x for x in network.values()][i]) for i in range(len(network))))
 
@@ -168,31 +170,54 @@ def compute_network(
         if reach["seqorder"] not in ordered_reaches:
             ordered_reaches.update(
                 {reach["seqorder"]: []}
-            )  # TODO: Should this be a set/dictionary?
+            )  
         ordered_reaches[reach["seqorder"]].append([head_segment, reach])
 
     # initialize flowdepthvel dict
-    nts = 50  # one timestep
-    # nts = 1440 # number fof timestep = 1140 * 60(model timestep) = 86400 = day
+    # nts = 50  # one timestep
+    nts = 143  # test with dt =10
+    dt = 300  # in seconds
+    # nts = 1440 # number of  timestep = 1140 * 60(model timestep) = 86400 = day
+
+    # initialize write to files variable
+    writeToCSV = True
+    writeToNETCDF = True
+    pathToOutputFile = os.path.join(root, "test", "output", "text") 
 
     for ts in range(0, nts):
-        # print(f'timestep: {ts}\n')
-
         for x in range(network["maximum_reach_seqorder"], -1, -1):
             for head_segment, reach in ordered_reaches[x]:
-                # print(f'{{{head_segment}}}:{reach}')
 
                 compute_mc_reach_up2down(
                     head_segment=head_segment,
                     reach=reach,
                     supernetwork_data=supernetwork_data,
                     ts=ts,
+                    dt=dt, 
                     verbose=verbose,
                     debuglevel=debuglevel,
                     write_output=write_output,
                     assume_short_ts=assume_short_ts,
                 )
-                # print(f'{head_segment} {flowdepthvel[head_segment]}')
+
+    if (writeToCSV):
+        for x in range(network['maximum_reach_seqorder'], -1, -1):
+            for head_segment, reach in ordered_reaches[x]:
+                printarray(reach=reach
+                           , verbose=verbose
+                           , debuglevel=debuglevel
+                           , pathToOutputFile=pathToOutputFile
+                           )
+
+    if (writeToNETCDF):
+        writeArraytoNC(network=network
+                       , nts=nts
+                       , dt=dt
+                       , verbose=verbose
+                       , debuglevel=debuglevel
+                       , pathToOutputFile=pathToOutputFile
+                       )
+
 
 
 # TODO: generalize with a direction flag
@@ -201,6 +226,7 @@ def compute_mc_reach_up2down(
     reach=None,
     supernetwork_data=None,
     ts=0,
+    dt=60, 
     verbose=False,
     debuglevel=0,
     write_output=False,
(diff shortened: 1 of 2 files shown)


src/python_routing_v01/compute_nhd_routing_SingleSeg.py
```diff
@@ -453,22 +615,39 @@ def main():
...
+    # Lateral flow
+    ## test 1. Take lateral flow from wrf-hydro output from Pocono Basin
+    ql_input_folder = os.path.join(
+        root, r"test/input/geo/PoconoSampleData2/Pocono_ql_testsamp1_nwm_mc.csv"
+    )
+    ql = pd.read_csv(ql_input_folder, index_col=0)
+
+    for index, row in ql.iterrows():
+        # print(index, row.to_numpy())
+        flowdepthvel[index]["qlatval"] = row.to_numpy().tolist()
```
groutr, line 638: Replace `row.to_numpy().tolist()` with `row.to_list()`

jameshalgren: We saw this and we'll address it in the final PR. 
@jdhondia-noaa

jameshalgren: Updated in [80765c6](https://github.com/NOAA-OWP/t-route/pull/77/commits/80765c6c259688f461298076fa9eff4753a9307f)

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
```diff
@@ -453,40 +697,60 @@ def main():
...
+        run_pocono_test
+    ):  # test 1. Take lateral flow from wrf-hydro output from Pocono Basin
+        ql_input_folder = os.path.join(
+            root, r"test/input/geo/PoconoSampleData2/Pocono_ql_testsamp1_nwm_mc.csv"
+        )
+        ql = pd.read_csv(ql_input_folder, index_col=0)
+
+    else:
+        q = np.full((len(connections), nts), qlat_const, dtype="float32")
+        ql = pd.DataFrame(q, index=connections.keys(), columns=range(nts))
```
groutr, line 727: Pandas has this use case covered.
```python
ql = pd.DataFrame(qlat_const, index=connections.keys(), columns=range(nts))
```
No need to allocate the numpy array first.

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
```diff
@@ -453,40 +697,60 @@ def main():
...
+
+    for index, row in ql.iterrows():
+        flowveldepth[index]["qlatval"] = row.tolist()
```
groutr, line 730: This needs to be `row.to_list()`. `row` is a pandas Series object.

groutr: This should have raised an exception during runtime.

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
```diff
@@ -302,52 +322,256 @@ def compute_mc_reach_up2down(
...
+                    if debuglevel <= -2:
+                        print(f"{current_segment} (tail)")
+                    break
+                next_segment = connections[current_segment]["downstream"]
+                if debuglevel <= -3:
+                    print(f"{current_segment} --> {next_segment}\n")
+                current_segment = next_segment
+
+    # check number of timesteps should match the time the data is written
+    if int(len(flowveldepth_data["time"][0])) != int(nts):
```
groutr, line 472: `len` always returns an int. We don't need to cast here.

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
```diff
@@ -302,52 +322,256 @@ def compute_mc_reach_up2down(
...
+# arguments reach , pathToOutputFile
+# using global connections and flowveldepth.
+def printarray(
+    reach=None, verbose=False, debuglevel=0, pathToOutputFile="../../test/output/text"
+):
+    global connections
+    global flowveldepth
+
+    # define CSV file Header
+    header = [["time", "qlat", "q", "v", "d"]]
```
groutr, line 366: Why is this a nested list?

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
```diff
@@ -302,52 +322,256 @@ def compute_mc_reach_up2down(
...
+
+
+# ### Psuedocode
+# Write Array  to CSV file
+# arguments reach , pathToOutputFile
+# using global connections and flowveldepth.
+def printarray(
+    reach=None, verbose=False, debuglevel=0, pathToOutputFile="../../test/output/text"
+):
+    global connections
```
groutr, line 362: These global statements are only needed if we are intending to mutate them. They are not needed if we only read from them.

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
```diff
@@ -86,15 +121,17 @@ def _handle_args():
...
 
     return parser.parse_args()
 
 
-ENV_IS_CL = False
-if ENV_IS_CL:
-    root = "/content/wrf_hydro_nwm_public/trunk/NDHMS/dynamic_channel_routing/"
-elif not ENV_IS_CL:
-    root = os.path.dirname(os.path.dirname((os.path.abspath(""))))
+root = pathlib.Path("../..").resolve()
```
groutr, line 134: Refer to https://github.com/NOAA-OWP/t-route/issues/78#issuecomment-657788810
Do `pathlib.Path(__file__).parent.parent.resolve()`

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
jameshalgren: They also help with the parallel processing -- as long as we run on Linux, which uses copy-on-write, then the global dataset  is never duplicated when we pass it to the various versions of the parallel execution call. That was the deeper idea behind having it that way in the first place.

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
jameshalgren: fixed. (not sure if I did that or @jdhondia-noaa ...)

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
jameshalgren: Because the writerows call expects a nested iterable. If you give it the strings, it iterates them: 
```
"t","i","m","e"
"q","l","a","t"
"q"
"v"
"d"
"0.0","10.0","0.8481932878494263","0.04473819211125374","0.010925279930233955"
```

groutr: Sorry, I meant to leave a comment below to use `csvwriter.writeheader()`

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
jameshalgren: @jdhondia-noaa was being very careful. 

nts (number of time steps) should always be an integer, too, though. 

Really, if they are not equal, we definitely want an error, even if the error arises from strange typing, because it would be just as much of a problem, so I removed the casting.

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
jameshalgren: But -- since the globals inside the parallel call are being used in a nested function (to print results for the different networks, I think the functional use actually makes sense. Some simple tests seemed to show that it doesn't mess up the parallel optimization, so I've made this change. We can consider going back fairly easily, if needed. Here's the diff: 

```
diff --git a/src/python_routing_v01/compute_nhd_routing_SingleSeg.py b/src/python_routing_v01/compute_nhd_routing_SingleSeg.py
index f81fc8c..d9eb4b2 100644
--- a/src/python_routing_v01/compute_nhd_routing_SingleSeg.py
+++ b/src/python_routing_v01/compute_nhd_routing_SingleSeg.py
@@ -224,7 +224,9 @@ def compute_network(
     if writeToCSV:
         for x in range(network["maximum_reach_seqorder"], -1, -1):
             for head_segment, reach in ordered_reaches[x]:
-                printarray(
+                writeArraytoCSV(
+                    connections=connections,
+                    flowveldepth=flowveldepth,
                     reach=reach,
                     verbose=verbose,
                     debuglevel=debuglevel,
@@ -233,6 +235,8 @@ def compute_network(

     if writeToNETCDF:
         writeArraytoNC(
+            connections=connections,
+            flowveldepth=flowveldepth,
             network=network,
             nts=nts,
             dt=dt,
@@ -356,11 +360,14 @@ def compute_mc_reach_up2down(
 # Write Array  to CSV file
 # arguments reach , pathToOutputFile
 # using global connections and flowveldepth.
-def printarray(
-    reach=None, verbose=False, debuglevel=0, pathToOutputFile="../../test/output/text"
+def writeArraytoCSV(
+    flowveldepth=None,
+    connections=None,
+    reach=None,
+    verbose=False,
+    debuglevel=0,
+    pathToOutputFile="../../test/output/text",
 ):
-    global connections
-    global flowveldepth

     # define CSV file Header
     header = [["time", "qlat", "q", "v", "d"]]
@@ -401,6 +408,8 @@ def printarray(
 # arguments network, number of timesteps (nts),  timestep in seconds  (dt),  pathToOutputFile
 # using global connections and flowveldepth.
 def writeArraytoNC(
+    flowveldepth=None,
+    connections=None,
     network=None,
     nts=0,
     dt=60,
@@ -408,8 +417,6 @@ def writeArraytoNC(
     debuglevel=0,
     pathToOutputFile="../../test/output/text/",
 ):
-    global connections
-    global flowveldepth
     # create  array variables to copy from python "flowveldepth" which is global
     flowveldepth_data = {
         "segment": [],
```

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
jameshalgren: Ok. That makes more sense. Making that change...

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
jameshalgren: it looks like it is `row.to_list()`... must have been fixed...

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
jameshalgren: fixed per comment.

jameshalgren:
@jdhondia-noaa 
I am going to accept this and suggest that we now make a branch called short-time-step to work on the issues with that particular feature.

PR #90, opened 2020-07-21
Adding level-pool reservoir simulation capability to v01 

jameshalgren:
This adds major functionality to v01
- WRF-Hydro direct data reading
- NetCDF output
- Corrected Short-Time-Step toggling
- generalized JSON input control
- additional output data (storage, qlatCum)

The pertinent concepts would need to be ported to v02, and we can freeze v01.


src/python_framework_v01/nhd_network_utilities_v01.py
```diff
@@ -380,6 +382,19 @@ def set_supernetwork_data(
             "title_string": "Pocono Test Example",
             "driver_string": "ESRI Shapefile",
             "layer_string": 0,
+            "water_body_parameter_file_type": "Level_Pool",
+            "level_pool_waterbody_parameter_file_path": os.path.join(
+                geo_input_folder, "NWM_2.1_Sample_Datasets", "LAKEPARM_CONUS.nc"
+            ),
+            "level_pool_waterbody_area": "LkArea",  # area of reservoir
+            "level_pool_weir_elevation": "WeirE",
+            "level_pool_waterbody_max_elevation": "LkMxE",
```
groutr, line 391: I would strongly suggest moving all the "level_pool*" entries to a nested dictionary.

jameshalgren: That makes sense -- these dictionaries are mirrored in the json custom input file format, which ultimately should be the way we provide input to the operational model. A nested dictionary should work with that seamlessly.

src/python_framework_v01/nhd_network_utilities_v01.py
```diff
@@ -638,3 +669,128 @@ def set_networks(supernetwork="", geo_input_folder=None, verbose=True, debugleve
...
+    """
+
+    xds = xr.open_dataset(crosswalk_file)
+
+    xdf = xds.to_dataframe()
+    xdf = xdf.reset_index()
+    xdf = xdf[[channel_ID_column]]
+
+    qds = xr.open_dataset(channel_initial_states_file)
+    qdf = qds.to_dataframe()
```
groutr, line 725: We should get avoid this dataframe creation. We are essentially reading the entire file into memory here when we only only want to read a couple of variables.
```python
qdf = qds[[us_flow_column, ds_flow_column, depth_column]].to_dataframe()
```

src/python_framework_v01/nhd_network_utilities_v01.py
```diff
@@ -638,3 +669,128 @@ def set_networks(supernetwork="", geo_input_folder=None, verbose=True, debugleve
...
+    us_flow_column: column in the restart file to use for upstream flow initial state
+    ds_flow_column: column in the restart file to use for downstream flow initial state
+    depth_column: column in the restart file to use for depth initial state
+    crosswalk_file: File containing channel IDs IN THE ORDER of the Restart File
+    channel_ID_column: field in the crosswalk file to assign as the index of the restart values
+    """
+
+    xds = xr.open_dataset(crosswalk_file)
+
+    xdf = xds.to_dataframe()
```
groutr, line 720: Replace this with
```python
xdf = xds[channel_ID_column].to_dataframe()
```

src/python_framework_v01/nhd_network_utilities_v01.py
```diff
@@ -638,3 +669,128 @@ def set_networks(supernetwork="", geo_input_folder=None, verbose=True, debugleve
...
+    qdf.rename(
+        columns={
+            us_flow_column: default_us_flow_column,
+            ds_flow_column: default_ds_flow_column,
+            depth_column: default_depth_column,
+        },
+        inplace=True,
+    )
+
+    mod = xdf.join(qdf)
```
groutr, line 742: If we create the dataframes as above, we no longer need to join.
We can simply add the link column (pandas will align the indexes automatically).

src/fortran_routing/mc_pylink_v00/Reservoir_singleTS/module_levelpool.f90
```diff
@@ -0,0 +1,237 @@
...
+! sub-objects. This module also contains the
+! subroutine to run level pool reservoir that is
+! derived from the reservoir base type interface
+! to run reservoir. Running level pool will
+! then call the LEVELPOOL_PHYSICS subroutine, which
+! processes the given inputs, properties, and
+! state for a particular level pool reservoir and
+! returns the output/outflow.
+
+module module_levelpool
```
jameshalgren, line 22: @kmsampson
Kevin, 
I'm tagging you on the PR into our routing repository... hopefully it's not too confusing. I thought you might be interested. We used the f2py wrapper to build a python call directly to the level pool function from the WRF-Hydro code. We've made a few minor modifications such as specifying the precision (because f2py is not careful with that in the translation) and making the depth variable more explicit (in - H0, out - H1). Eventually, we will use a custom cython wrapper so we can operate outside the GIL.

test/input/json/CustomInput.json
```diff
@@ -1,21 +1,78 @@
...
-    "topwidth_col": 11,
-    "topwidthcc_col": 12,
-    "MusK_col": 7,
-    "MusX_col": 8,
-    "ChSlp_col": 12,
-    "terminal_code": 0,
-    "driver_string": "ESRI Shapefile",
-    "layer_string": 0
+{   
+    "run_parameters":
```
jameshalgren, line 2: @groutr 
The files checked in in this PR now include this demo file and everything needed to execute a full simulation: 
1. lateral inflows
2. reservoirs
3. initial conditions in the reservoirs
4. initial conditions in the channels

jameshalgren: (Pointing that out ... I'll be working soon on porting things over to v02)

jameshalgren:
I've confirmed that our script still produces the output for the canonical test from Dong Ha when using the short-time-step option. As we know, there is work still to confirm the behavior of the model when the short-time-step option is not in use, however, we will work on that under another issue/pull-request.

Any approving reviewer, when you have a moment, please approve the request and I will merge it into the repository.

jdhondia-noaa:
ready to be merged and test

PR #107, opened 2020-08-28
Use pre-declared numpy arrays for flowveldepth in v01

jhrehanoaa:
Replaced singleseg file with updated numpy array file.


src/python_routing_v01/compute_nhd_routing_SingleSeg.py
```diff
@@ -597,10 +597,10 @@ def compute_mc_reach_up2down(
         # storage
         volumec = dt * (quc - qdc + qlat)
         # TODO: This qlatCum is invalid as a cumulative value unless time is factored in
-        qlatCum = qlat * dt
+        qlatCum = qlat
```
jameshalgren, line 600: we need to leave this line with dt

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
```diff
@@ -254,7 +254,7 @@ def in_wsl() -> bool:
     WSL is thought to be the only common Linux kernel with Microsoft in the name.
     """
 
-    return "microsoft" in (os.uname().release).lower()
```
jameshalgren, line 257: need to leave this line with the `.lower()`

jhrehanoaa: i pushed two commits that change these, but I can't see it updated in files changed. Only in the commits tab.

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
```diff
@@ -347,13 +347,13 @@ def compute_network(
...
     flowveldepth = {
         connection: {
-            "time": [],
-            "qlatCumval": [],
-            "qlatval": [],
-            "flowval": [],
-            "velval": [],
-            "depthval": [],
-            "storageval": [],
+            "time": np.zeros(nts),
```
groutr, line 350: An alternative idea would be to use a single numpy array with 2 dimensions. This would have the added benefit that since the array is contiguous, you could update all the values in a timestep in a single operation. I would suggest a (nts, 7) shaped array.
Then your update operation is simply
```python
flowveldepth[current_segment][ts] = [ts*dt, qlatCum, qlat, qdc, velc, depthc, volumec]
```

groutr: columns 0-6 are time, qlatCumval, qlatval, flowval, velval, depthval, storageval respectively. The row corresponds to the timestep

src/python_routing_v01/compute_nhd_routing_SingleSeg.py
```diff
@@ -743,18 +743,18 @@ def compute_level_pool_reach_up2down(
 
     volumec = dt * (quc - qdc + qlat)
     # TODO: This qlatCum is invalid as a cumulative value unless time is factored in
-    qlatCum = qlat * dt
+    qlatCum = qlat
```
jameshalgren, line 746: We need to preserve line 746. 
``` qlatCum = qlat * dt ```

the  `* dt` is important.

Comment (jhrehanoaa) [reply src/python_routing_v01/compute_nhd_routing_SingleSeg.py:746]: I will try to put in those changes. This should have been changed in one of my last commits. I will double-check why it did not update.

File: src/python_routing_v01/compute_nhd_routing_SingleSeg.py (hunk 3 above)
Comment (jameshalgren) [reply src/python_routing_v01/compute_nhd_routing_SingleSeg.py:350]: Is there any benefit in access speed to having it be (nts, 7) vs. (7, nts)? (i.e., having the timesteps as the column dimension?)
I can see that the single-line assignment would be more complicated...

Comment (groutr) [reply src/python_routing_v01/compute_nhd_routing_SingleSeg.py:350]: It boils down to C-contiguous vs Fortran-contiguous. Numpy arrays are by default C-contiguous so a shape of (nts, 7) means that the values for each time step are contiguous. If you have a (7, nts) array, unless the array is Fortran-contiguous, the values for the timestep are strided across the array. Performance wise, it would be very difficult to pin down the effect because of so much else going on, but it can make a difference.

Comment (groutr) [reply src/python_routing_v01/compute_nhd_routing_SingleSeg.py:350]: Here are some timings:
```
%timeit C[500]  # shape=(1440, 7) C-contig. The resulting view is contiguous.
179ns
%timeit C1[:, 500] #shape (7, 1440) C-contig.  The resulting view is noncontiguous.
381ns
%timeit F[:, 500] # shape (7, 1440) F-contig. Resulting view is contiguous
277ns
%timeit F1[500] #shape (1440, 7), F-contig. Resulting view is noncontiguous.
197ns
```

jhrehanoaa: Copying this over to a new pull request. We will be merging the other changes in the meantime.

jameshalgren: I'm trying to digest the timing information. We'll take this in two steps -- Let's accept this PR to use the arrays, the reshape them in a second piece of work.

PR #111, opened 2020-08-31
Fix AREAC and WPC initialization; remove velp as input

jameshalgren:
Fix AREAC and WPC initialization; remove velp as input …

Also removed duplicate secant function
Add additional comment blocks with code
to return to previous initialization capability
for compatibility with current WRF-Hydro code
see NCAR/wrf_hydro_nwm_public#510
for the same concept in the main WRF-Hydro code

Update single seg routing for no velp; 
Update singlesegment MC test script for new initialization

Add test at high flow to demonstrate capability of compound channel
Also added function built from direct extract from WRF-Hydro for comparison.


awlostowski-noaa:
complete this PR with a Courant output capability @awlostowski-noaa

awlostowski-noaa (changes requested):
Need to change https://github.com/NOAA-OWP/t-route/blob/master/notebooks/mc-wave-routing-demo.ipynb accordingly. Else, this notebook will no longer run as an illustrative example. Other notebooks deprecated?

jameshalgren:
> Need to change https://github.com/NOAA-OWP/t-route/blob/master/notebooks/mc-wave-routing-demo.ipynb accordingly. Else, this notebook will no longer run as an illustrative example. Other notebooks deprecated?

@awlostowski-noaa  The `singlesegment` function called by the notebook has been adapted to use the new module, so now no change is required in the notebook. The function arguments remain unchanged as well, for the moment, so any instances where the singlesegment function from the demo script was applied, as in the notebook, there should be no change.

jameshalgren:
> complete this PR with a Courant output capability @awlostowski-noaa

Added new issue #111

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/mc_sseg_stime_NOLOOP_demo.py
```diff
@@ -62,8 +131,8 @@ def singlesegment(
     n_manning_cc=None,  # ncc
     cs=None,  # cs
     s0=None,  # s0
-    velp=None,  # velocity at previous time step
     depthp=None,  # depth at previous time step
+    velp=None,  # DUMMY -- dropped from the computation
```
awlostowski-noaa, line 135: OK! I was initially worried that the velp variable would be one too many, but it is hardwired to NONE - so there should be no problems. I am going to test the mc-wave-routing-demo in Colab from this branch to be sure.

awlostowski-noaa: Notebook will indeed run just fine. Confirmed with test in Colab from this branch

awlostowski-noaa (dismissed):

awlostowski-noaa:
@awlostowski-noaa check to make sure this doesn't affect v02 routing

awlostowski-noaa:
I created a [Colab notebook to test the compatibility of the WPC initialization fix on the v02 routing framework](https://colab.research.google.com/drive/1ffv0tRgIZEsNMUC6PY9WyOlaqVfgoRM7?authuser=1#scrollTo=BKE0JwQD-jPK). After cloning this branch, the notebook shows that cython modules are successfully compiled and the code is executable. 

That said, I think we are good to go forward with a merge!

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/MCsingleSegStime_f2py_NOLOOP.f90
```diff
@@ -6,7 +6,7 @@ module muskingcunge_module
 contains
 
 subroutine muskingcungenwm(dt, qup, quc, qdp, ql, dx, bw, tw, twcc,&
-    n, ncc, cs, s0, velp, depthp, qdc, velc, depthc)
+    n, ncc, cs, s0, depthp, velp, qdc, velc, depthc)
```
awlostowski-noaa, line 9: A note on why this commit is necessary:

Including `velp` as an argument is necessary, even though it not actually used. The `muskingcungenwm` subroutine is called in many pieces of code in the repo, for example, [`mc_reach.compute_reach_kernel`](https://github.com/NOAA-OWP/t-route/blob/master/src/python_routing_v02/fast_reach/mc_reach.pyx#L51). If we were to delete `velp` it would change the number of arguments immediately break lots of things.

awlostowski-noaa (approved):

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/MCsingleSegStime_f2py_NOLOOP.f90
awlostowski-noaa: @jameshalgren Actually - I just noticed that the order of these arguments has been switched from velp, depthp to depthp, velp. This matters? 

This might be an insidious bug because the model will run b/c the inputs structure is correct, though the outputs may be garbage.

awlostowski-noaa (approved):

jameshalgren (commented):
Done.

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/MCsingleSegStime_f2py_NOLOOP.f90
jameshalgren: I think that may have been an old display. I swapped that order back -- this is how it looks now: 
``` fortran
subroutine muskingcungenwm(dt, qup, quc, qdp, ql, dx, bw, tw, twcc,&
    n, ncc, cs, s0, velp, depthp, qdc, velc, depthc)
```

awlostowski-noaa (approved):

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/mc_sseg_stime_NOLOOP_demo.py
jameshalgren: Noted. Thanks.

PR #114, opened 2020-09-02
Owp progress bar

jhrehanoaa:
added command line and json level reach/segment progress completion for SingleSeg. 

There is an error happening when I try to run this with the custominput json file. It claims one of reach ids being computed does not exist in the list I created from the networks object.  "ValueError: 4185667 is not in list" is what it returns when I run the custominput file.


jameshalgren:
Let's use tqdm.

jameshalgren:
Replaced (fixing conflicts, catching bugs -- thanks @jhrehanoaa!) with #147 
Closing.

PR #129, opened 2020-09-14
Catch up blackening

jameshalgren:
FORMATTING ONLY


jameshalgren:
@awlostowski-noaa @groutr Pushing this formatting only commit.

PR #130, opened 2020-09-15
Clarify explanation of short-time-step

jameshalgren:
Comment only; no code change. 

Clarifying the distinction between the time-step sequence and the topology sequence.


jameshalgren:
tag @jhrehanoaa

jhrehanoaa (approved):

PR #142, opened 2020-09-18
v01 Warmstate function updates

jameshalgren:
Several minor updates to the v01 warmstate functions (improving docstrings, removing unused function inputs, adding appropriate default values)

Also added a functional call to create a dataframe for the csv-based qlateral data files.


awlostowski-noaa:
let's move new functions in to v02 and call from v01

awlostowski-noaa:
@awlostowski-noaa put qlateral function capability in this PR.

jhrehanoaa (approved):

jhrehanoaa (changes requested):
additional changes being added

jhrehanoaa (dismissed):

src/python_framework_v02/nhd_io.py
```diff
@@ -55,9 +47,202 @@ def replace_downstreams(data, downstream_col, terminal_code):
...
+    no capability for handling non-uniform timesteps in the qlaterals.
+
+    The qlateral may also be input using comma delimited file -- see
+    `get_ql_from_csv`
+    """
+
+    li = []
+
+    for filename in qlat_files:
+        ds = xr.open_dataset(filename)
```
groutr, line 119: There should be a `ds.close()` somewhere here.

jameshalgren: Converted to `with` style

jameshalgren:
> @awlostowski-noaa put qlateral function capability in this PR.

Moved all these read functions to nhd_io.

src/python_framework_v02/nhd_io.py
```diff
@@ -55,9 +48,211 @@ def replace_downstreams(data, downstream_col, terminal_code):
...
+
+    The qlateral may also be input using comma delimited file -- see
+    `get_ql_from_csv`
+    """
+
+    li = []
+
+    for filename in qlat_files:
+        with xr.open_dataset(filename) as ds:
+            df1 = ds.to_dataframe()
```
awlostowski-noaa, line 130: call nhd_io.read_netcdf ? Apparent redundancy

awlostowski-noaa (approved):
changes look good. Good call moving functions to nhd_io. Should facilitate transition to v2.

src/python_framework_v02/nhd_io.py
```diff
@@ -55,9 +48,211 @@ def replace_downstreams(data, downstream_col, terminal_code):
...
+
+
+def get_ql_from_wrf_hydro(qlat_files, index_col="station_id", value_col="q_lateral"):
+    """
+    qlat_files: globbed list of CHRTOUT files containing desired lateral inflows
+    index_col: column/field in the CHRTOUT files with the segment/link id
+    value_col: column/field in the CHRTOUT files with the lateral inflow value
+
+    In general the CHRTOUT files contain one value per time step. At present, there is
+    no capability for handling non-uniform timesteps in the qlaterals.
```
awlostowski-noaa, line 120: Is this something that needs action to be operational? If so create a new issue?

jameshalgren: Right now, there will never be a non-uniform qlateral timestep. What you have started to look at with reading the actual time-stamps heads in the direction of being able to handle that. The core computation engine should probably never have to be fully date aware, but I'm not sure. In any case, the comment is a warning and reminder for post MVP development, I think.

src/python_framework_v02/nhd_io.py
jameshalgren: Good point. Currently, the `nhd_io.read_netcdf` is identical to the lines you indicated in the `nhd_io.get_ql_from_wrf_hydro` function. I had seen that, but prompted by your comment, I looked more carefully at a number of different ways to combine the functions and to include, perhaps, some of the other similar calls to the xarray library. However, every other place that we call the `xarray.read_dataset()` function, there is some sort of filtering applied immediately to the dataset  before generating the dataframe and these are relatively specific in each instance (some have preconditions on the filter, some swap indexes, etc.) 

So... I think we should leave it for now to avoid the potentially confusing suggestion of a connection between the the two functions other than the fact that the same xarray code is called. At some later point, we might take on a broader project to refine the io capabilities.

PR #153, opened 2020-09-25
[WIP - Hold] Building a reach-threaded parallel test

jameshalgren:
This is a work in progress __do not merge__

The first goal here is to execute a simulation that threads reaches using either Just-In-Time or Opportunistic parallelization. See readme.md file in this PR for additional explanation. 

At this point in the work in progress, we have a first level method for producing the thread order for parallelization and now are working to establish the execution.


jameshalgren:
@groutr -- code as WIP

src/python_framework_v02/nhd_network.py
```diff
@@ -274,12 +273,106 @@ def dfs_decomposition_depth2(N, path_func, source_nodes=None):
...
+    Returns:
+        load_dict
+    """
+
+    load_dict = {}
+
+    for t in tuple_list:
+        k = t[key_idx]
+        v = t[val_idx]
+        load_dict.setdefault(k, []).append(v)
```
groutr, line 373: This use case is what defaultdict was created for. You might prefer defaultdict here.
```python
load_dict = defaultdict(list)
for t in tuple_list:
    load_dict[t[key_idx]].append(t[val_idx])
```
The advantage of defaultdict over setdefault is that setdefault is constructing a list for every lookup into load_dict.

jameshalgren: Good point. I converted that over for this function.

jameshalgren:
Threaded parallelization has been achieved through other work.

PR #154, opened 2020-10-01
[WIP  - Hold] Implement/Update logging methods

jhrehanoaa:
__NOTE:__ Holding off on further development here; need to rebase and re-approach this issue from the perspective of v02. 

still amending a few things, but a basic concept.  Will add more print outs for each function that seems reasonable without flooding the output. This can be stored in a log file as well.


src/python_routing_v02/compute_nhd_routing_SingleSeg_v02.py
```diff
@@ -193,15 +194,15 @@ def main():
...
     if showtiming:
-        print("... in %s seconds." % (time.time() - start_time))
+        logger.warning("... in %s seconds." % (time.time() - start_time))
 
     # STEP 2
     if showtiming:
         start_time = time.time()
     if verbose:
-        print("organizing connections into reaches ...")
+        logger.warning("organizing connections into reaches ...")
```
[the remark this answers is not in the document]
jhrehanoaa: I'll make those changes. I just left the print outs from --verbose as warning print outs to stay consistent.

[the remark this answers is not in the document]
groutr: I see. I would also suggest removing the checks for verbose. The logging library controls verbosity by setting the log level. The verbose flag should be determining the log level and nothing else.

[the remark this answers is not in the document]
jhrehanoaa: Okay, would it be beneficial to remove the debug-level as well and only specify the log level from the command line arguments?  It seems pointless to have both of them when this should replace it.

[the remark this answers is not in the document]
groutr: You can still have the user specify the debug-level and verbosity and use that to determine what log level should be. After that is done, there is no need to do any further checking.
When you log a message, the very first thing that the library checks is the logging level. It will only continue to process messages that are greater than or equal to the log level. This removes the need for checking in your application code as you're offloading the work to the logging library.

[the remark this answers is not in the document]
groutr: See: https://docs.python.org/3/howto/logging.html#when-to-use-logging

src/python_framework_v02/nhd_network.py
```diff
@@ -2,17 +2,20 @@
 from itertools import chain
 from functools import reduce, partial
 from collections.abc import Iterable
+import logging
 
+LOG = logging.getLogger(__name__)
 
 def nodes(N):
     yield from N.keys() | (v for v in chain.from_iterable(N.values()) if v not in N)
+    LOG.debug("Edges: %s", N)
```
groutr, line 11: Note that N can be be quite large. Printing the network might not be a very useful log message. A perhaps, more useful log message might be the expected number of nodes in N `len(N) + sum(map(len, N.values()))`

src/python_framework_v02/nhd_network.py
```diff
@@ -2,17 +2,20 @@
...
 
 
 def edges(N):
     for i, v in N.items():
         for j in v:
             yield (i, j)
-
+    LOG.debug("Edges: %s", N)
```
groutr, line 18: See comment in `nodes()`

src/python_framework_v02/nhd_network.py
```diff
@@ -27,6 +30,7 @@ def in_degrees(N):
     degs = Counter(chain.from_iterable(N.values()))
     degs.update(dict.fromkeys(headwaters(N), 0))
     return degs
+    LOG.debug("In_degrees: %s", N)
```
groutr, line 33: This line will never execute because of the return statement.

src/python_framework_v02/nhd_network.py
```diff
@@ -278,13 +283,13 @@ def reservoir_shore(connections, waterbody_nodes):
     for node in wbody_set:
         shore.update(filter(not_in, connections[node]))
     return list(shore)
-
+    LOG.debug("reservoir_shore - connections: %s", connections, "waterbody_nodes: %s", waterbody_nodes)
```
groutr, line 286: This will likely raise an error. We're using the older string formatting language so it would look more like:
`LOG.debug("reservoir_shore - connections: %s waterbody_nodes: %s", connections, waterbody_nodes)`

jhrehanoaa:
Question regarding logging: The old network utilities and networkbuilder type files were passed verbose, debuglevel to them in order to sometimes generate print statements. Is there a simple way anyone would suggest that would allow the set logging level in compute_nhd_connections to easily be passed to the entire networkbuilder or utilities file? It seems to me that I would need to pass the logging level to each function used in those files the same way debuglevel and verbose is now. Looking for a workaround to this that doesn't require adding this to each function while still preserving the log file identification specified in each file. 

Ex.)  from comput_nhd_routing_Sseg.py

    if supernetwork_parameters:
        supernetwork_values = nnu.get_nhd_connections(
            supernetwork_parameters=supernetwork_parameters,
            verbose=False,
            debuglevel=debuglevel,
        )

We see the verbose and debuglevels being communicated to nnu. However, with the logger I would need to know the verbose and debuglevel ahead of time to set the level within the nnu file unless each function is being passed it.

    if supernetwork_parameters:
        supernetwork_values = nnu.get_nhd_connections(
            supernetwork_parameters=supernetwork_parameters,
            verbose=False,
            debuglevel=debuglevel,
            LOG=LOG
        )

Something similar to this etc. 

I didn't know if there was a workaround for this or an idea on how best to approach this? I'd like to still maintain a log location of each file name in the output so errors can be tracked. Here is some current output returned from compute_nhd_routing_Sseg. However, the utility file and others should replace the 'SingleSeg' name.


2020-10-15 15:55:40,740 - SingleSeg - INFO - ordered reach computation complete
2020-10-15 15:55:40,740 - SingleSeg - INFO - ... in 0.9463725090026855 seconds.
2020-10-15 15:55:40,740 - SingleSeg - INFO - program complete
2020-10-15 15:55:40,740 - SingleSeg - INFO - ... in 1.4401917457580566 seconds.


If it is easiest to simply pass the log level information through each function I can do that. It just seems incorrect to handle it that way. It would be nice to set the level in compute_nhd_routing_Sseg and flag on/off if you want it to also output from the utility and other files. You can't really run the utility and networkbuilders in a standalone debug without the inputs coming from compute_nhd_routing_Sseg or else it doesn't help when something actually goes wrong.

groutr:
@jhrehanoaa It is usually entirely unnecessary to pass around the log level. Loggers form a tree a tree and if a loglevel isn't explicitly defined in a node, the parent propagates its loglevel to the child, up till a loglevel is defined. Typically, it's only necessary to configure to root logger.

>Loggers have a concept of effective level. If a level is not explicitly set on a logger, the level of its parent is used instead as its effective level. If the parent has no explicit level set, its parent is examined, and so on - all ancestors are searched until an explicitly set level is found. The root logger always has an explicit level set (WARNING by default). When deciding whether to process an event, the effective level of the logger is used to determine whether the event is passed to the logger’s handlers.

jhrehanoaa:
If it is unnecessary to pass it between files everything should be in order now with the correct naming convention. I believe it should essentially be tiered now as -v = 1(info), -v --debuglevel 1 = 2(debuglevel), -v --debuglevel 2 = 3(warning), -v --debuglevel 3 = 4(critical). 
I appreciate you weighing in on all of this.

src/python_routing_v02/compute_nhd_routing_SingleSeg_v02.py
```diff
@@ -112,6 +127,16 @@ def _handle_args():
 import mc_reach
 import nhd_network
 import nhd_io
+import set_logger as sl
+
+args = _handle_args()
```
jameshalgren, line 132: We are calling this twice --- needed?

src/python_framework_v01/networkbuilder.py
```diff
@@ -13,12 +16,13 @@ def get_down_connections(
     data_key=r"data",
     verbose=False,
     debuglevel=0,
+    LOG=LOG,
 ):
     # TODO: Consider moving debug and verbose prints to the calling function
     if debuglevel <= -100:
         breakpoint()
     if verbose:
```
jameshalgren, line 24: Do we still need these if statements with the LOG logic?

src/python_framework_v01/networkbuilder.py
```diff
@@ -273,11 +279,12 @@ def get_up_connections(
     downstream_key=r"downstream",
     verbose=False,
     debuglevel=0,
+    LOG=LOG,
```
jameshalgren, line 282: I understood that we don't need this explicit function argument here (i.e., ~~`LOG = LOG`~~)... will you check on that?

jameshalgren:
We've come to the place we are pretty organically. Thinking about what we would want to support, here are some scenarios I can think of: 
1) Nothing. No logging output at all; Only the numerical program results. 
2) Some logging to file; no output to screen
3) Some logging to file; different output to screen
4) Same output to file and to screen
5) No file logging; some output to screen

There are also some special cases that are more like tests, but we can separate those. E.g, the --pretend or the --network-only switches. 

Default behavior

Turning on the logging flag should use this
```
    parser.add_argument(
        "--parallel",
        nargs="?",
        help="Use the parallel computation engine (omit flag for serial computation)",
        dest="parallel_compute",
        const="type3",
    )

File: src/python_framework_v01/networkbuilder.py (hunk 8 above)
Comment (groutr) [reply src/python_framework_v01/networkbuilder.py:282]: That's correct. You don't need to pass the log around like that. You can either use a module level log or you can get a handle to the log via `getLogger`

**Comment by jameshalgren** [comment]
For the log append mode, I suggest we assume overwrite, then provide a flag something like this: 
```
     parser.add_argument(
        "--log_file_append",
        help="Append log messages to log file (instead of overwriting, which is the default behavior.)",
        dest="log_append_mode",
        action="store_true",
    )

```

**Comment by jhrehanoaa** [comment]
I tried switching out the if verbose or if debuglevel lines. Sadly, it just prints every statement regardless of the debuglevel. Am I improperly setting the debuglevel at the beginning? 

`if verbose:
    LOG.info("identifying upstream connections and junctions ...")`

to 

`LOG.info("identifying upstream connections and junctions ...")`

The expected behavior should just be logging when info is the set LOG level. However, it looks like regardless it will log all messages indiscriminate of what log level is set originally. Will continue to investigate, but I thought the .setlevel command had already accomplished this.

**Comment by groutr** [comment]
@jhrehanoaa I think your `set_logger()` is incorrect.  I think you should draw out a diagram of your logging tree and understand what exactly you are setting on each log.

**Comment by jhrehanoaa** [comment]
From what I can see online it looks like the logger is set, but instead reads all messages above that level not below. 

https://www.geeksforgeeks.org/logging-in-python/            - has an example of the numerical values associated to each level. 
I can rework the ordering from info->critical to critical->info and remove the IF statements. Otherwise, I think the IF statements keep this working for what levels we assign. I wasn't sure if critical being printed for basic prints statements would be more helpful than keeping it as is. Thoughts on which direction to proceed with that?

**Comment by groutr** [comment]
@jhrehanoaa that's correct. Levels are inclusive. So if INFO is set, you'll also get WARNING, ERROR, and CRITICAL messages. Your debug ordering didn't make sense to me and that is why I suggested this more common pattern I often see. (https://github.com/NOAA-OWP/t-route/issues/144#issuecomment-710056999).

The suggestion in that comment maps nicely to the concept of levels. Each time -v is given, it drops the level by 10 (or 1 log level). 
```
prog.py  # This sets logging to WARNING (the default for logging). Presumably, these are the really important messages that users should see because they may affect the validity of the output.
prog.py -v # Set logging to INFO. User can see messages about normal operation of the code.
prog.py -v -v # Set logging to DEBUG. User sees more detail about operation of the code.
prog.py --quiet # Set logging level to ERROR/CRITICAL (silencing warnings). These are messages that the user must see.
```
I would, at the very least, reverse the ordering of your debuglevel. Intuitively, it doesn't make sense that a higher debuglevel results in less verbosity. I would expect that increasing debuglevel increases the amount of output.

I still think you need to revisit your `set_logger` function. The compute_nhd* files need to be setting up the root logger, not their own loggers.

jameshalgren:
> I would, at the very least, reverse the ordering of your debuglevel. Intuitively, it doesn't make sense that a higher debuglevel results in less verbosity. I would expect that increasing debuglevel increases the amount of output.

The idea in the original debug levels/verbose was something like this: 
Verbose -- print human readable titles to tell where we are in the console program. These are most useful during an interactive execution and might actually get in the way of a non-interactive run. 
Debug level 1 -- print counts of network analysis
Debug level 2 -- print counts of network analysis + junction and network info
Debug level 3 -- 1+2+ print output of everything possible so that we can look at the network analysis in detail and try to figure out if something was missed. This level is really only useful for the very small cases where we know all of the output we expect, because beyond that the output is so large as to be incomprehensible. 

I'm having a little difficulty meshing that picture with the idea of error logging where the information level is tied to the severity of an error where we aren't even talking about error trapping in what we prepared to this point...

I've moved this to draft/WIP status for the moment to make sure we're doing what we really mean to do.

groutr:
The way that I understand things are mapped right now:
Verbose - INFO (20)
Debug level 1 - DEBUG(10)
Debug level 2 - WARNING (30)
Debug level 3 - CRITICAL (50)

The left side means to ascend in verbosity, while the left side is definitely decreasing in verbosity. That is what I was pointing out. (Note the "discontinuity" at debug level 1)

The way things _should_ be mapped are:
Verbose - INFO (20)
Debug level 1 - DEBUG (10)
Debug level 2 - (a value _less than_ debug level 1) Call this DEBUG_EXTRA
Debug level 3 - (a value _less than_ debug level 2) Call this DEBUG_ALL

We need to define two custom log levels: DEBUG_EXTRA = 9, DEBUG_ALL=8.
If debug level is 3, and messages we emit at that level are visible for all logs with a level >= 8.

Logging isn't just for errors. Logging is a reporting mechanism where we can easily tune the verbosity. I would highly suggest scanning thru https://docs.python.org/3/howto/logging.html#basic-logging-tutorial. Most of the issues in this thread are answered in this tutorial.

jameshalgren:
This thread could use a log level :) I've added some more thoughts; we'll get it. As a reminder, with #160, we need to rebase this whole PR and it might be easier to just rebuild it, given what we're talking about here. Thanks @jhrehanoaa and @groutr for keeping this going.
 
> The way that I understand things are mapped right now:
> Verbose - INFO (20)
> Debug level 1 - DEBUG(10)
> Debug level 2 - WARNING (30)
> Debug level 3 - CRITICAL (50)
> 
> The left side means to ascend in verbosity, while the left side is definitely decreasing in verbosity. That is what I was pointing out. (Note the "discontinuity" at debug level 1)

Agreed. That's confusing. Both because we turned it upside down, at least partly, and also because the words are not consistent with the nature of the messages we have associated with the different levels. For instance, there is nothing 'Critical' about the deep debug output we're trying to get; there is no warning of any substantive nature in the second level (or in any other level for that matter -- we probably need another issue entirely about true error trapping and logging of that nature.)

> 
> The way things _should_ be mapped are:
> Verbose - INFO (20)
> Debug level 1 - DEBUG (10)
> Debug level 2 - (a value _less than_ debug level 1) Call this DEBUG_EXTRA
> Debug level 3 - (a value _less than_ debug level 2) Call this DEBUG_ALL
> 
> We need to define two custom log levels: DEBUG_EXTRA = 9, DEBUG_ALL=8.
> If debug level is 3, any messages we emit at that level are visible for all logs with a level >= 8.

I like the `DEBUG_EXTRA` and `DEBUG_ALL` labels. It would make sense for those to sit on top of WARNING and CRITICAL (if we had any such messages.) Maybe the things we've discussed as Verbose should be simple print statements to the console. 

The way we're talking about it here, Verbose would be something like `DEBUG_MINI`. 

> 
> Logging isn't just for errors. Logging is a reporting mechanism where we can easily tune the verbosity. I would highly suggest scanning thru https://docs.python.org/3/howto/logging.html#basic-logging-tutorial. Most of the issues in this thread are answered in this tutorial.

That reference says that the simplest form of a log message is just a print statement -- would it make sense to keep a few of those for what we have been calling `verbose`? As I mentioned in a prior comment, there's a distinction between some sort of messaging to indicate progress of the program and a potentially more (or less?) detailed report in a log file.

groutr:
The names WARNING and CRITICAL are generic names that correspond to a number. I think it's more helpful to think of things in terms of their numeric value here. If one were to omit the verbose flag, what would the output consist of?

jameshalgren:
@jhrehanoaa -- we should consider pulling this material over to v02 -- we have an opportunity to redo logging and output functions a bit since we are moving to a new engine.

jhrehanoaa:
we could try, but it will require an overhaul of all the debuglevel and verbosity. We had issues before with the log level being passed between different files correctly though.

PR #158, opened 2020-10-14
[WIP] PR comparison tool

jhrehanoaa:
Looking for some feedback on this one. 

I have created the bones for something that can locate and compare v01 and v02 routing outputs from two different versions of the code. You can compare your master copy versus new code to quickly define if the outputs are equivalent still. It switches input directories automatically, you just need to provide the initial folder locations in the directory path for these to locate. Uses -m -n for arguments.

 I will be adding each of the debug levels to them etc. I can add any other file that is generally used or whatever output we need to compare against. I think this will make it easier to check PRs. Let me know your thoughts.


awlostowski-noaa:
Perhaps we consider putting this in a Jupyter notebook. When reviewing a PR, the reviewer can open the notebook via Google Colab, commands in the notebook can clone the master and PR repos and compare output. 

cell 1: Clone master repo, generate v01 and v02 results, hold results in memory
cell 2: remove the t-route directory and clone the PR repo, generate v01 and v02 results, hold results in memory
cell 3: compare results between master and PR.

groutr:
@awlostowski-noaa is it not possible to simply pull the remote branch instead of cloning the repo twice? You would then only need to clone once, pull in the remote branch, and then switch branches.

jhrehanoaa:
I can try changing this to switch between two different remote branches. Currently, it should be working now. 
You can enter 👍 
`python3 compare_files.py -m PR1_folder -n PR2_folder` 
 to select each pull request folder. It will run through the outputs of verbose, debuglevel 1, 2, and 3. 
It will print results like this: 

> PASS - results of compute_nhd_routing_SingleSeg.py match : debuglevel 0
> PASS - results of compute_nhd_routing_SingleSeg_v02.py match: debuglevel 0
> PASS - results of compute_nhd_routing_SingleSeg.py match : debuglevel 1
> PASS - results of compute_nhd_routing_SingleSeg_v02.py match: debuglevel 1
> PASS - results of compute_nhd_routing_SingleSeg.py match : debuglevel 2
> PASS - results of compute_nhd_routing_SingleSeg_v02.py match: debuglevel 2
> PASS - results of compute_nhd_routing_SingleSeg.py match : debuglevel 3
> PASS - results of compute_nhd_routing_SingleSeg_v02.py match: debuglevel 3
> (jupyter) [jacob.hreha@nwcal-ihcb-bm01 python_routing_v01]$ python3 compare_files.py 
> FAIL - results of compute_nhd_routing_SingleSeg.py are not equivalent : debuglevel 0
> PASS - results of compute_nhd_routing_SingleSeg_v02.py match: debuglevel 0
> FAIL - results of compute_nhd_routing_SingleSeg.py are not equivalent : debuglevel 1
> PASS - results of compute_nhd_routing_SingleSeg_v02.py match: debuglevel 1
> FAIL - results of compute_nhd_routing_SingleSeg.py are not equivalent : debuglevel 2
> PASS - results of compute_nhd_routing_SingleSeg_v02.py match: debuglevel 2
> FAIL - results of compute_nhd_routing_SingleSeg.py are not equivalent : debuglevel 3
> PASS - results of compute_nhd_routing_SingleSeg_v02.py match: debuglevel 3


I will modify the code to do a remote branch check.

jameshalgren:
I'm seeing us head in two directions at once here:
1) Facilitating comparison of any code to some standard result -- something more along the lines of a unit test.
2) Facilitating comparison between branches of our code.

I would say that our first check should be against a standard result. We could add an option to check, from within a modified branch, against results of some other branch. But I can see how having a versioned script move around the current git working copy could yield some strange results. 

So... I would suggest we get a test against a standard output set up first, then experiment with the idea of dynamic comparisons. @awlostowski-noaa @groutr @jhrehanoaa .... thoughts?

jameshalgren:
Moving this capability to v01 technical debt. Discussions around Output capability may lead us back to these commits. Or they could be superseded. Tagging @jhrehanoaa for situational awareness.

jameshalgren:
Closing. Functionality may be used later.

PR #161, opened 2020-10-23
Building a reach-threaded parallel test (2.0)

awlostowski-noaa:
This is a work in progress do not merge

This pull request builds on #153 

The new reach ordering designed by @jameshalgren is implemented on the most recent initial conditions enabled version of mc_reach.pyx. A parity test is conducted to ensure that the new reach ordering scheme produces identical output to the original reach ordering scheme, and ultimately WRF hydro routing results. 

Once we are confident that the new reach ordering scheme is functionally equivalent to the original v02 formulation, we will experiment with multithreading reaches of in the same reach ordering group. 

For additional context - see #153. Here, @jameshalgren has documented his thinking in designing a strategy for reach-threaded parallelization of v02.


awlostowski-noaa:
For WRF parity, see [pocono1_warmstart_v2.ipynb](https://github.com/awlostowski-noaa/t-route/blob/v2_ordering/notebooks/pocono1_warmstart_v2.ipynb)

This notebook should prove that the new ordering scheme works.

awlostowski-noaa:
# Update
1. A new function has been added to [`mc_reach.pyx`](https://github.com/awlostowski-noaa/t-route/blob/v2_ordering/src/python_routing_v02/fast_reach/mc_reach.pyx), named `compute_network_multithread`. This function produces the same result as `compute_network` but does so using a multithreaded approach based on a new reach grouping concept. 
2. [`pocono1_warmstart_v2.ipynb`](https://github.com/awlostowski-noaa/t-route/blob/v2_ordering/notebooks/pocono1_warmstart_v2.ipynb) has been updated to show that the `compute_network_multithread` and `compute_network` functions produce identical output and *very* nearly reproduce WRF Hydro routing results. 
3. `nhd_network_utilities_v02.py` has been updated with the current (v2.0.4) version of NWM.

awlostowski-noaa:
# Suggested Review Actions
1. Run [`pocono1_warmstart_v2.ipynb`](https://github.com/awlostowski-noaa/t-route/blob/v2_ordering/notebooks/pocono1_warmstart_v2.ipynb) via Google Colab to prove to yourself that the multithreaded version of `compute_network` works as it is intended. 
2. Look at `mc_reach.compute_network_multithread` to see how it differes from the original `mc_reach.compute_network`. Are there any efficiencies we can capitalize on? Any variables that are initialized but unused? Is the overhead too burdensome? 
3. Suggestion welcome on how to fix broken `compute-network-v2.ipynb` notebook.

awlostowski-noaa:
## BEFORE MERGE...
Make sure to change the git clone commands in the demo notebook

awlostowski-noaa:
# PROFILE needed
A quick look at applying the multithreaded compute_network on a large network shows that it is actually much SLOWER (?!) than the original. This is likely because I have introduced an unacceptable amount of overhead. I would appreciate some help in profiling the multithreaded function to see why it takes much longer than expected.

jameshalgren (approved):

PR #162, opened 2020-10-30
Build support

hellkite500:
Add setup.py for building/install modules required for fast_reach, and some (still somewhat manual) support for building the required fortran sources.

## Changes

- change the import pattern for `reach` components in `mc_reach`

## Testing

```sh
cd t-route/src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS
make install
cd t-route/src/python_routing_v02
```
Then `python setup.py install  --use-cython` or `python setup.py install` to install the required extension modules.

Tested using `python compute_nhd_routing_SingleSeg_v02.py -n Pocono_TEST1`

## Checklist

- [x] Linux
- [x] MaxOs


hellkite500:
@groutr couldn't tag you as a reviewer, but please take a look if you get a chance.

jameshalgren:
Committed this on accident? src/python_routing_v02/fast_reach/mc_reach.c

hellkite500:
> Committed this on accident? src/python_routing_v02/fast_reach/mc_reach.c

No, that was intentional.  This way you don't need cython to build and install the module.  These are static source files, and should be considered part of the distribution.  Now I could be convinced to remove it from the repo itself and put it only in a distribution artifact, but I wasn't thinking that at the time.

groutr:
@hellkite500 I would probably leave the generated c files out of the repo. I think it is less clutter to only include them in the distribution artifacts.

src/python_routing_v02/fast_reach/mc_reach.pyx
```diff
@@ -7,8 +7,12 @@ from numpy cimport ndarray
 cimport numpy as np
 cimport cython
 
-from reach cimport muskingcunge, QVD
-
+#NJF For whatever reason, when cimporting muskingcunge from reach, the linker breaks in weird ways
```
groutr, line 10: Odd, I don't recall having any trouble with that.

src/python_routing_v02/setup.py
```diff
@@ -0,0 +1,42 @@
+from distutils.core import setup
+from distutils.extension import Extension
+import sys
+import numpy as np
+
+"""
+If source ships with the cython generated .c code, then cython isn't a hard requirement
+This setup can use if if passed the --use-cython flag, otherwise it looks for the c source
```
groutr, line 8: I wonder if a more robust logic would be to check for cython first, then, if cython isn't callable, check for the c files. That way, we can issue a warning if the version of cython is too old. I think we require cython >= 0.29.1, but recommend the latest version. This also means that we use freshly generated c files.

hellkite500: That is actually part of the problem, once the C code is generated, the cython version doesn't matter, the extension is built based on the static C code.  If you re-cythonize each time you distribute, then different versions of cython may produce different C code which may or may not work.

The way it exists now gives someone the ability to use cython and rebuild if they know what they are doing, but for all other cases, it just builds the extension from the existing C code.

I am going to remove the .c files from this PR FOR NOW until the development gets a little more complete.

hellkite500:
Updated to remove generated .c files, create a namespace package for the framework modules, and add some extension types to the framework.

To use, run
```sh
cd src/python_framework_v02/
python setup.py install --use-cython
cd src/python_routing_v02/
python setup.py install --use-cython
python compute_nhd_routing_SingleSeg_v02.py
```

src/python_framework_v02/troute/network/reach.pxd
```diff
@@ -0,0 +1,25 @@
+cimport numpy as np
```
jameshalgren, line 1: I think we can drop the .pxd; and __init__.py

I ran a test with an empty troute folder in the python site-packages and no .pxd (nor __init__.py for that matter) and was able to install without trouble.

groutr: What do you mean by "drop the .pxd"?

jameshalgren:
> Updated to remove generated .c files, create a namespace package for the framework modules, and add some extension types to the framework.
> 
> To use, run
> 
> ```shell
> cd src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS
> make all
> make install
> 
> cd src/python_framework_v02/
> python setup.py install --use-cython
> 
> cd src/python_routing_v02/
> python setup.py install --use-cython
> 
> python compute_nhd_routing_SingleSeg_v02.py
> ```

Created #163 to remind us to revisit this.

jameshalgren (approved):

PR #166, opened 2020-11-12
(Outdated -- use #175) build subnetworks

awlostowski-noaa:
A new utility to build subnetworks of common order from larger networks.

### DIFF
--- a/src/python_framework_v02/nhd_network.py
+++ b/src/python_framework_v02/nhd_network.py
@@ -418,7 +418,7 @@ def replace_waterbodies_connections(connections, waterbodies):
             new_conn[n] = connections[n]
     return new_conn
 
-def build_subnetworks(connections, rconn, max_depth, sources=None):
+def build_subnetworks(connections, rconn, min_size, sources=None):
     '''
     Construct subnetworks using a truncated breadth-first-search
     
@@ -455,6 +455,7 @@ def build_subnetworks(connections, rconn, max_depth, sources=None):
                 reachable_depth = set()
                 Q = deque([h])
                 D = deque([0])
+                stop_depth = 1000000
                 while Q or D:
 
                     x = Q.popleft() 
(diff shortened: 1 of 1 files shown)


awlostowski-noaa:
# suggested improvement
Currently, subnetwork creation is constrained by a user-appointed maximum subnetwork depth. Subnetworks are created via a breadth-first-search. The search begins from a source node and stops once the maximum depth is reached OR once there are no more nodes to reach.

**Should we change slightly such that the search continues until either 1) a threshold number of segments are in the subnetwork or 2) there are no more nodes to reach?** The issue with the code as it stands now, is that the depth of the subnetwork is not immediately related to the size (number of segments) in the subnetwork.

src/python_framework_v02/nhd_network.py
```diff
@@ -417,3 +417,85 @@ def replace_waterbodies_connections(connections, waterbodies):
...
+                    reachable.add(x)
+                    reachable_depth.add(y)
+
+                    if len(rconn.get(x, ())) > 1:
+                        us_depth = y+1
+                    else:
+                        us_depth = y
+
+                    if us_depth <= max_depth:
+                        D.extend([us_depth]*len(rconn.get(x, ())))
```
jameshalgren, line 471: Need additional criteria for hitting a minimum number of segments aggregated into the group.

jameshalgren (dismissed):

jameshalgren:
Replaced by #175

src/python_framework_v02/nhd_network.py
jameshalgren: I think we have this functionality as it now stands in PR #175 (built from this PR code) So I'll resolve this conversation.

jameshalgren:
As noted, closing to use #175 which contains this code rebased to the current master + additional capability for using the grouped/clustered subnetworks in the actual routing function.

PR #178, opened 2020-12-03
Include short time step arg in compute_network call

jameshalgren:
Short-time-step argument was not included in the script calls to the compute_network function. 

This PR corrects the omission.

NOTE: Initial testing shows that the output is consistent with our Pocono_TEST2 standard test. The computational error will be corrected in a separate PR.


jhrehanoaa (approved):

jameshalgren:
Tag @awlostowski-noaa @groutr for situational awareness.

PR #175, opened 2020-11-27
Add subnetwork calculation options

jameshalgren:
The idea of parallelizing by reaches is attractive but the opportunity for parallelism is more than offset by the in-efficiency of reducing the calculation to such small chunks as are found in individual reaches. This PR provides initial code to break the network into groups of subordinate reaches, "subnetworks". Only preliminary attempts at connecting these subnetworks have been made in the PR as it stands now -- more changes will be needed in the `compute_network` code to allow for additional values in the `flowveldepth` array, from which the upstream inflow values are obtained to provide boundary conditions for each call to the `compute_reach_kernel`. The subnetworks are passed to the compute subnetwork function just-in-time, so that upstream subnetworks are compute just prior to computation of their downstream dependencies, hence, the commandline flag is `by-subnetwork-jit`

The following command-line demonstrates the call for parallel execution with subnetworks using subnets of at least 2 segments each:
 ipython compute_nhd_routing_SingleSeg_v02.py -- -v --debuglevel 1 --parallel by-subnetwork-jit --subnet-size 2 --nts 2

A second option, `by-subnetwork-jit-clustered`, builds on the previous idea, but also attempts to group onto a single thread subnetworks that are very small -- this becomes very important in the full-resolution routing networks. @awlostowski-noaa -- we might see if we can include here a few of the distribution plots.

PR #168, opened 2020-11-17
Bindings and Tests to Reservoir Module

jdmattern-noaa:
Cython to Fortran bindings to Reservoir module and Python unit tests of binding calls for each type of Reservoir

This PR is just for viewing and review purposes. We will need to coordinate on where to best integrate these bindings.

More documentation of modules and functions to follow. 

Main files to review for bindings and tests:
src/fortran_routing/Reservoir_Binding/kernels/test_compute_kernel.py

src/fortran_routing/Reservoir_Binding/kernels/compute_kernel_lp.pyx
src/fortran_routing/Reservoir_Binding/kernels/compute_kernel_hybrid.pyx
src/fortran_routing/Reservoir_Binding/kernels/compute_kernel_rfc.pyx 

src/fortran_routing/Reservoir_Binding/kernels/fortran/Reservoirs/bind_lp.f90 
src/fortran_routing/Reservoir_Binding/kernels/fortran/Reservoirs/bind_hybrid.f90
src/fortran_routing/Reservoir_Binding/kernels/fortran/Reservoirs/bind_rfc.f90 

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [ ] Code follows project standards (link if applicable)
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [ ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Target Environment support

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS
- [ ] Flexible from small to large screens
- [ ] No linting errors or warnings
- [ ] JavaScript tests are passing

PR #185, opened 2020-12-14
Model version 2.0 tag alpha

jameshalgren:
This branch has three updates that are headed towards a full version 2.0 of the model. 
#### What is here:
1) ability to process custom inputs
2) a re-factored main script with greatly simplified input coding and ability to import the functional call into other applications.
3) a new test script that currently contains a comparison test to show parity with WRF-Hydro output. 

#### To be added to complete version 2.0 (but not necessarily in this PR. 
1) Reservoir handling (see #76) (see also #168)
2) Subnetwork Parallel computation (see #175)
3) Data Assimilation handling (no PR yet -- @jhrehanoaa please update this comment with your PR # when you have the time-slice reading set up.)

### Test capability
This pull request includes a simple comparison between t-route and WRF-simulated flows for the Pocono1 Test case. When the Pocono1 test is run in compute_nhd_network_SingleSeg_v02.py the output will be automatically checked against WRF hydro channel flow simulations.

### Separate functional calls out of main
The `main()` function of the compute_nhd_routing_SingleSeg.py contains a data entry portion and a true computational portion. This PR separates those two and creates a `compute_nhd_routing_SingleSeg_v02` function that can be called with a dictionary of the basic parameters to execute the routing function. 

With this function separated, the script may be importing into a notebook and the input can be prepared and passed from a notebook, for instance, for checking and demonstrating parity.

A simple test of this PR is accomplished by executing
```
python compute_nhd_routing_SingleSeg_v02.py -f ../../test/input/json/CustomInput.json
```


src/python_framework_v02/troute/nhd_network_augment.py
```diff
@@ -716,128 +746,156 @@ def main():
...
     # update RouteLink data
-    RouteLink_edit = RouteLink.loc[data_merged.index.values]    
+    RouteLink_edit = RouteLink.loc[data_merged.index.values]
 
-    for (columnName, columnData) in data_merged.iteritems(): 
-        RouteLink_edit.loc[:,columnName] = columnData
+    for (columnName, columnData) in data_merged.iteritems():
+        RouteLink_edit.loc[:, columnName] = columnData
 
     for idx in RouteLink_edit.index:
```
groutr, line 836: An operation like this is much better expressed in pandas as:
```python
RouteLink_edit["to"] = RouteLink_edit["to"].clip(lower=0)
```
We avoid the explicit iterating thru the index which will be slow.

jameshalgren: @awlostowski-noaa Please make a note of this as an issue and we can address it in a separate PR.

awlostowski-noaa (approved):

jameshalgren:
@jmccreight

PR #149, opened 2020-09-22
Yaml changes merge

jhrehanoaa:
fixed the merge changes in this PR. Should work fine.

PR #188, opened 2021-01-11
Subnetwork final + Model Version 2.0 alpha2

jameshalgren:
__TL; DR:__ Major Update -- basically all features of T-Route V 2.0 except for Reservoirs, Data Assimilation, and Output control. Test features of this PR with the following commands: 
- micro test of all features: `python compute_nhd_routing_SingleSeg_v02.py -f ../../test/input/yaml/CustomInput.yaml`; 
- a bigger dataset with no WRF-Hydro input: `python compute_nhd_routing_SingleSeg_v02.py -n Mainstems_CONUS -tv --nts 288 --parallel by-subnetwork-jit-clustered --subnet-size 150 --debuglevel 1`
- __New Issue/PR:__ Add Falls Lake Florence Event (requires not-yet-available input datasets + input file) 
- __New Issue/PR:__ Add CONUS Benchmark Test (currently only works on Cheyenne; requires input file)
<br>__Important__ This PR Replaces #166 and #175 and is the follow-on for #185
# Summary
This PR includes 
1. __a new function `build_subnetworks`__ that takes the subnet creation function of the 'subnetwork_creation' branch and pushes those subnetworks through a threaded call to the compute_network script. A number of upgrades implemented here reduce the most critical performance bottlenecks so that CONUS Full Resolution simulation is now feasible.  
2. Ability to execute based on an __input .json or .yaml file defining all other input parameters__ or data file locations. As inputs and model controls become more complex, this ability permits more efficient input 
3. Integrated capability for __reading channel warm states from WRF-Hydro restart files__. Code for reservoir warm states is present but should be brought on-line and tested in another PR. 
4. Integrated capability for __comparing the t-route output to the Wrf-Hydro output__ for the same set of input qlaterals and restart files. This builds on capability introduced as part of #185 and is a key tool for demonstrating parity with existing NWM functions.
5. Reorganization of the v02 execution and compile scripts (organization, dev option for package install, blacking, additional cli options...)

This is the next level of the version 2.0 _alpha_ of our framework (see #185 initial alpha). When the reservoir and Data Assimilation capabilities are incorporated with the capabilities in this PR, that should represent the full _beta_. Delivered as such, with any appropriate testing and minor performance or feature upgrades, the framework will be able to replace all relevant functions of the existing channel routing capability within the NWM framework. In addition to reservoir and DA integration, standardization of outputs to match the current NWM, and incorporation of more appropriate logging are two important elements of the final delivery.

# Explanation
## 1) build_subnetworks 

With this PR, four execution modes are availble: `serial` (default), `--parallel by-network`; `--parallel by-subnetwork-jit`; and `--parallel by-subnetwork-jit-clustered`

Tests:
```
python compute_nhd_routing_SingleSeg_v02.py -n Mainstems_CONUS -tv --nts 288 --debuglevel 1
python compute_nhd_routing_SingleSeg_v02.py -n Mainstems_CONUS -tv --nts 288 --parallel --debuglevel 1
python compute_nhd_routing_SingleSeg_v02.py -n Mainstems_CONUS -tv --nts 288 --parallel by-subnetwork-jit --subnet-size 150 --debuglevel 1
python compute_nhd_routing_SingleSeg_v02.py -n Mainstems_CONUS -tv --nts 288 --parallel by-subnetwork-jit-clustered --subnet-size 150 --debuglevel 1
```
On a workstation with 36 cores (which are all used by default under the parallel scheme with no `--cpu-pool` option specified), this runs in 92, 42, 24, and 24 seconds, respectively -- not a full linear scaling, but a clear improvement. These simulations use a constant qlateral at all segments and a warm state of 0.

__New Issue/PR:__ The output shown with the debuglevel is sorted differently for the serial vs. parallel tests above... fix this. <br>

The scaling is sensitive to the subnet-size. Choosing a too-large subnet-size reduces the potential parallelizability of the problem; a too-small subnet-size incurs significant overhead because the GIL must be reacquired, as presently configured, in between each level of subnetwork.

__New Issue/PR:__ Encapsulate the entire build_subnetwork capability into a function -- currently, the output requires further processing before it can be used in the `compute_network` routing engine.<br>

The test has also been executed with the `CONUS_FULL_RES_v20` network. Execution on the full-resolution dataset exposes an important limitation of our subnetwork development process. When the network is extremely dense (as in the case of the full-resolution NHD-based NWM hydrofabric), the subnetwork generation algorithm produces many small subnetworks at the 'leaves' of the traversal (i.e., near the headwaters). If these are simulated individually, they use the parallel pool resources very inefficiently. The workaround presented here is to cluster the small subnetworks together and include them with other subnetworks so that the computational load is more evenly distributed across threads. While there are a number of ways this could be improved, my recommendation is that we look at two in the near term (see below.)
```
# Standard serial execution
  
# Standard by-network parallelization
ipython compute_nhd_routing_SingleSeg_v02.py -- -t -n CONUS_FULL_RES_v20 -v --debuglevel 1 --parallel --nts 100 --cpu-pool 12
_JIT Preprocessing time_ 0 seconds.
_PARALLEL TIME = total time_  seconds

# Just-in-time Subnetwork parallelization _with no_ clustering
ipython compute_nhd_routing_SingleSeg_v02.py -- -t -n CONUS_FULL_RES_v20 -v --debuglevel 1 --parallel by-subnetwork-jit --subnet-size 15000 --nts 100 --cpu-pool 12
_JIT Preprocessing time_  seconds.
_PARALLEL TIME_  seconds.
_Total time_  seconds.

# Just-in-time Subnetwork parallelization _with_ clustering
ipython compute_nhd_routing_SingleSeg_v02.py -- -t -n CONUS_FULL_RES_v20 -v --debuglevel 1 --parallel by-subnetwork-jit-clustered --subnet-size 15000 --nts 100 --cpu-pool 12
_JIT Preprocessing time_  seconds.
_PARALLEL TIME_  seconds.
_Total time_  seconds.
```

### Just-In-Time parallelization
The "Just-in-Time" or _JIT_ idea is explained further in our [AGU poster](https://agu2020fallmeeting-agu.ipostersessions.com/Default.aspx?s=3B-5C-F9-F6-F6-5F-E3-AC-C6-38-7F-8B-14-48-F7-37&pdfprint=true&guestview=true): 

__New Issue/PR:__ Get the AGU content into a Markdown file in this Git repository. (Also consider publishing via ORCID.)

* Calculate first the headwater reaches of edges of the longest network...
* Followed by all reaches below headwaters, etc.
* Orchestrating the computation so each dependency is computed just before it is needed downstream (hence "Just-in-time") provides best theoretical potential speedup.
* Practical efficiencies are obtained by grouping the reaches into cascading orders of subnetworks.

image
_Caption: Grouping the reaches into subnetworks balances the practical impact of many parallel calls. An optimal subnetwork size is small enough to permit sufficient parallelism and large enough to ensure that parallel overhead is not burdensome. Colors denote subnetworks of common reverse network order. Higher order subnetworks are computed prior to lower order subnetworks in order to maintain topological dependencies._

### Opportunistic Parallelization
Using a very similar analysis of the network, but ordering instead from the headwaters, we could identify all the subnetworks that could execute simultaneously in a first pass, and so on down to the bottom. While this seems like a nuanced difference from the JIT (which calculates any given subnetwork only just before it is needed by the downstream neighbors; or in other words, the precedents to each subnetwork are guaranteed to be part of the calculation immediately preceding it), the Opportunistic parallelization approach with uniformly sized subnetworks actually guarantees a pretty significant bottleneck towards the end of a very large network (i.e., the Mississippi), when only a single series of reaches or subnetworks remains. Undoubtedly, there are approaches to laying out the subnetworks to minimize the difficulty... 

### Options for improving the subnetwork creation
Per comment above, these are two most likely candidates for improving the subnetwork algorithm: 
1) __New Issue/PR:__ Implement a subnetwork construction scheme that generates even-sized networks (might need heuristics?; expands on above TODO to compact/refactor build_subnetwork); 
2) __New Issue/PR:__ Experiment with and document computational performance impacts of thinning the network.

A number of other possibilities include the following:
- For parallel execution path between the network-based parallelization and the subnetwork-based parallelization, it seems like the the function we need to imitate is `reachable_network`, which produces the `independent_networks` object, as they are called in this branch. `reachable_network` calls the `reachable` function, which I propose could be adapted to accept option inputs to specify split points (__see note below__)
- The return from reachable_network could include additional "remainder" sets specifying the next upstream elements from what was reached in the BFS traversal -- if the traversal continues to the leaves of the network, the "remainder" set would be empty and if the traversal is split somehow, the remainder would be a set of new subnetwork tailwaters to continue tracing upstream.
- The split points could be specified by any of the following (not an exhaustive list of options):
    - reservoir inlets and outlets
    - a maximum number of segments or junctions (the tracing algorithm would have to step backward after passing the threshold)
    - a minimum number of segments or junctions (the tracing algorithm would have to continue one step forward after passing the threshold; __this is what is currently implemented in `build_subnetworks`.__
    - any arbitrary list of junction segments
    - a number of subnetworks to target (would require additional functions to count and divide into one of the more elemental subdivision types.)

_Also_... While not strictly necessary for execution, we are choosing to require that the subdivision must always split at junctions. This greatly simplifies the logic to connect between the cascading subnetworks.

## 2) Yaml/Json input control
Reading a custom input file was introduced incrementally in pull requests #149 #137 #90 and is moved forward an additional level with this PR.

A full (though very small) test of the input capability is available by running the CustomInput file:
```
#Identical output achieved with `python compute_nhd_routing_SingleSeg_v02.py -f ../../test/input/yaml/CustomInput.yaml`
python compute_nhd_routing_SingleSeg_v02.py -f ../../test/input/json/CustomInput.json
creating supernetwork connections set
supernetwork connections set complete
... in 0.09403371810913086 seconds.
organizing connections into reaches ...
reach organization complete
... in 0.0011286735534667969 seconds.
setting channel initial states ...
channel initial states complete
... in 0.027464866638183594 seconds.
creating qlateral array ...
qlateral array complete
... in 0.6489253044128418 seconds.
executing routing computation ...
           (0, q)    (0, v)    (0, d)    (1, q)    (1, v)    (1, d)    (2, q)    (2, v)    (2, d)    (3, q)  ...  (284, d)  (285, q)  (285, v)  (285, d)  (286, q)  (286, v)  (286, d)  (287, q)  (287, v)  (287, d)
4185209  0.012739  0.173687  0.041693  0.012738  0.173681  0.041691  0.012737  0.173677  0.041689  0.012736  ...  0.041409  0.012595  0.172934  0.041409  0.012595  0.172933  0.041409  0.012595  0.172933  0.041409
4185221  0.037830  0.254202  0.062983  0.037826  0.254193  0.062979  0.037823  0.254185  0.062976  0.037820  ...  0.062349  0.037196  0.252580  0.062348  0.037195  0.252577  0.062347  0.037194  0.252575  0.062346
4185223  0.002712  0.013621  0.165343  0.002712  0.013621  0.165340  0.002712  0.013621  0.165337  0.002712  ...  0.164396  0.002686  0.013578  0.164393  0.002686  0.013578  0.164390  0.002686  0.013578  0.164387
4185225  0.005960  0.077459  0.034619  0.005960  0.077458  0.034618  0.005959  0.077458  0.034618  0.005959  ...  0.034494  0.005924  0.077278  0.034494  0.005924  0.077277  0.034493  0.005924  0.077277  0.034493
4185231  0.015957  0.170060  0.029026  0.015957  0.170060  0.029026  0.015957  0.170060  0.029026  0.015957  ...  0.028984  0.015918  0.169898  0.028984  0.015918  0.169897  0.028983  0.015918  0.169896  0.028983
...           ...       ...       ...       ...       ...       ...       ...       ...       ...       ...  ...       ...       ...       ...       ...       ...       ...       ...       ...       ...       ...
4153682  0.023992  0.029227  0.585429  0.023991  0.023443  0.387796  0.023990  0.029227  0.585419  0.023989  ...  0.583662  0.023707  0.023357  0.385194  0.023706  0.029180  0.583649  0.023705  0.023356  0.385174
4153684  0.025567  0.180425  0.063529  0.025566  0.180423  0.063528  0.025565  0.180420  0.063526  0.025565  ...  0.063075  0.025262  0.179609  0.063073  0.025261  0.179605  0.063071  0.025260  0.179603  0.063070
4153688  0.009095  0.138854  0.042681  0.009094  0.138850  0.042679  0.009094  0.138847  0.042678  0.009093  ...  0.042333  0.008972  0.138133  0.042332  0.008972  0.138132  0.042331  0.008971  0.138130  0.042331
4153804  0.000872  0.166968  0.005594  0.000872  0.166956  0.005594  0.000872  0.166951  0.005594  0.000872  ...  0.005555  0.000863  0.166195  0.005555  0.000863  0.166195  0.005555  0.000863  0.166195  0.005555
4153806  0.004806  0.015872  0.212524  0.004805  0.015872  0.212517  0.004805  0.015872  0.212511  0.004805  ...  0.210802  0.004739  0.015802  0.210796  0.004738  0.015802  0.210791  0.004738  0.015802  0.210785

[336 rows x 864 columns]
ordered reach computation complete
... in 1.156254529953003 seconds.
conducting parity check, comparing WRF Hydro results against t-route results
                     flow, wrf (cms)  flow, t-route (cms)
2018-01-01 02:00:00         0.873438             0.873438
2018-01-01 03:00:00         0.873347             0.873347
2018-01-01 04:00:00         0.873250             0.873250
2018-01-01 05:00:00         0.873148             0.873148
2018-01-01 06:00:00         0.873040             0.873040
2018-01-01 07:00:00         0.872928             0.872928
2018-01-01 08:00:00         0.872810             0.872810
2018-01-01 09:00:00         0.872687             0.872687
2018-01-01 10:00:00         0.872559             0.872558
2018-01-01 11:00:00         0.872425             0.872425
2018-01-01 12:00:00         0.872288             0.872288
2018-01-01 13:00:00         0.872145             0.872145
2018-01-01 14:00:00         0.871998             0.871998
2018-01-01 15:00:00         0.871845             0.871845
2018-01-01 16:00:00         0.871689             0.871688
2018-01-01 17:00:00         0.871527             0.871527
2018-01-01 18:00:00         0.871361             0.871361
2018-01-01 19:00:00         0.871191             0.871191
2018-01-01 20:00:00         0.871015             0.871016
2018-01-01 21:00:00         0.870836             0.870837
2018-01-01 22:00:00         0.870653             0.870653
2018-01-01 23:00:00         0.870465             0.870465
2018-01-02 00:00:00         0.870273             0.870273
2018-01-02 01:00:00         0.870077             0.870077
```

__New Issue/PR:__ (maybe) a minor issue introduced by converting this to v02 format is that the custom input file format no longer works with the v01 script. If we wish, we could adjust the v01 scripts to use the named-columns. 

## 3) WRF-Hydro Warm States Read
The ability to ingest a warm state dataframe has been present for a while. This simply uses the already-built v01 function to read warm states and passes them into the proper dataframe in v02. 

__New Issue/PR:__ As visible in the above output from the CustomInput.json simulation, the is *almost* identical (values on 10, 16, 20, an 21 differ by .000001...). It would be good to be absolutely clear on the reason for the deviations since we have proven that correlation can be binary identical if the inputs are identical --- I suspect that there is a difference in the precision of the stored warm states vs. their internal value when used by WRF-Hydro. 

## 4) Test capability
The test capability is triggered by including the following elements in the custom input file (populated with the default test data in the repository): 
```
parity_parameters:
    parity_check_input_folder: "../../test/input/geo/NWM_2.1_Sample_Datasets/Pocono_TEST1/example_CHRTOUT/"
    parity_check_file_pattern_filter: "/*.CHRTOUT_DOMAIN1"
    parity_check_file_index_col: feature_id
    parity_check_file_value_col: streamflow
    parity_check_compare_node: 4186169
```

__New Issue/PR:__ Add identical command-line flags in a command group. 
__New Issue/PR:__ Rebuild the time-step flexible capabilities in the original --test pocono1 

## 5) Miscellaneous
This series of updates includes a number of refinements ultimately leading to being able to execute a CONUS-wide parity test and benchmark run. What we learned in heading that direction is that we have significant work to do to handle I/O more effectively. 

__New Issue/PR:__ Create self-looping capability to allow simulations of arbitrary length; for instance, the 8-year benchmark simulation would take quite a long time to load, if it loaded at all, and would likely far exceed any reasonable memory limit during execution.

### DIFF
--- a/src/python_framework_v02/troute/nhd_io.py
+++ b/src/python_framework_v02/troute/nhd_io.py
@@ -79,7 +79,7 @@ def replace_downstreams(data, downstream_col, terminal_code):
 def read_waterbody_df(waterbody_parameters, waterbodies_values, wbtype="level_pool"):
     """
     General waterbody dataframe reader. At present, only level-pool
-    capability exists. 
+    capability exists.
     """
     if wbtype == "level_pool":
         wb_params = waterbody_parameters[wbtype]
@@ -94,14 +94,14 @@ def read_level_pool_waterbody_df(
     parm_file, lake_index_field="lake_id", lake_id_mask=None
 ):
     """
-    Reads LAKEPARM file and prepares a dataframe, filtered 
+    Reads LAKEPARM file and prepares a dataframe, filtered
     to the relevant reservoirs, to provide the parameters
     for level-pool reservoir computation.
 
     Completely replaces the read_waterbody_df function from prior versions
     of the v02 routing code.
 
-    Prior version filtered the dataframe as opposed to the dataset as in this version. 
+    Prior version filtered the dataframe as opposed to the dataset as in this version.
     with xr.open_dataset(parm_file) as ds:
         df1 = ds.to_dataframe()
     df1 = df1.set_index(lake_index_field).sort_index(axis="index")
@@ -138,7 +138,9 @@ def read_qlat(path):
     return get_ql_from_csv(path)
 
 
-def get_ql_from_wrf_hydro_mf(qlat_files, index_col="station_id", value_col="q_lateral", filter_list=None):
+def get_ql_from_wrf_hydro_mf(
+    qlat_files, index_col="station_id", value_col="q_lateral", filter_list=None
+):
     """
     qlat_files: globbed list of CHRTOUT files containing desired lateral inflows
     index_col: column/field in the CHRTOUT files with the segment/link id
@@ -184,18 +186,18 @@ def get_ql_from_wrf_hydro_mf(qlat_files, index_col="station_id", value_col="q_la
     #       Assuming the bug is resolved at some point, the 'by_coords'
     #       method should be better for all cases.
     if len(qlat_files) == 1:
-        with xr.open_mfdataset(qlat_files, combine='nested', concat_dim='time') as ds:
+        with xr.open_mfdataset(qlat_files, combine="nested", concat_dim="time") as ds:
             if not filter_list:
                 df1 = ds[value_col].to_dataframe()
             else:
-                df1 = ds.sel({index_col:filter_list})[value_col].to_dataframe()
+                df1 = ds.sel({index_col: filter_list})[value_col].to_dataframe()
 
-    else: # Assumes > 1
-        with xr.open_mfdataset(qlat_files, combine='by_coords') as ds:
+    else:  # Assumes > 1
+        with xr.open_mfdataset(qlat_files, combine="by_coords") as ds:
             if not filter_list:
                 df1 = ds[value_col].to_dataframe()
             else:
-                df1 = ds.sel({index_col:filter_list})[value_col].to_dataframe()
+                df1 = ds.sel({index_col: filter_list})[value_col].to_dataframe()
 
     mod = df1.reset_index()
     ql = mod.pivot(index=index_col, columns="time", values=value_col)
--- a/src/python_framework_v02/troute/nhd_io.py
+++ b/src/python_framework_v02/troute/nhd_io.py
@@ -136,6 +136,64 @@ def read_qlat(path):
     return get_ql_from_csv(path)
 
 
+def get_ql_from_wrf_hydro_mf(qlat_files, index_col="station_id", value_col="q_lateral"):
+    """
+    qlat_files: globbed list of CHRTOUT files containing desired lateral inflows
+    index_col: column/field in the CHRTOUT files with the segment/link id
+    value_col: column/field in the CHRTOUT files with the lateral inflow value
+
+    In general the CHRTOUT files contain one value per time step. At present, there is
+    no capability for handling non-uniform timesteps in the qlaterals.
+
+    The qlateral may also be input using comma delimited file -- see
+    `get_ql_from_csv`
+
+
+    Note:
+    For later needs, filtering may be accomplished with one of:
+        ds.loc[{selectors}]
+        ds.sel({selectors})
+        ds.isel({selectors})
+
+    Returns from these selection functions are sub-datasets.
+
+    For example:
+    ```
+    (Pdb) ds.sel({"feature_id":[4186117],"time":ds.time.values[:2]})['q_lateral'].to_dataframe()
+                                     latitude  longitude  q_lateral
+    time                feature_id
+    2018-01-01 13:00:00 4186117     41.233807 -75.413895   0.006496
+    2018-01-02 00:00:00 4186117     41.233807 -75.413895   0.006460
+    ```
+
+    or...
+    ```
+    (Pdb) ds.sel({"feature_id":[4186117],"time":[np.datetime64('2018-01-01T13:00:00')]})['q_lateral'].to_dataframe()
+                                     latitude  longitude  q_lateral
+    time                feature_id
+    2018-01-01 13:00:00 4186117     41.233807 -75.413895   0.006496
+    ```
+    """
+
+    # TODO: compare performance with `combine='by_coords'`
+    # NOTE: This check for len(qlat_files) is because of a potential bug in
+    #       in open_mfdataset which strips the unit time dimension from an
+    #       input where only one timestamp (i.e., one input file) is provided.
+    #       Assuming the bug is resolved at some point, the 'by_coords'
+    #       method should be better for all cases.
+    if len(qlat_files) == 1:
+        with xr.open_mfdataset(qlat_files, combine='nested', concat_dim='time') as ds:
+            df1 = ds[value_col].to_dataframe()
+    else: # Assumes > 1
+        with xr.open_mfdataset(qlat_files, combine='by_coords') as ds:
+            df1 = ds[value_col].to_dataframe()
+
+    mod = df1.reset_index()
+    ql = mod.pivot(index=index_col, columns="time", values=value_col)
+
+    return ql
+
+
 def get_ql_from_wrf_hydro(qlat_files, index_col="station_id", value_col="q_lateral"):
     """
     qlat_files: globbed list of CHRTOUT files containing desired lateral inflows
--- a/src/python_framework_v02/troute/nhd_io.py
+++ b/src/python_framework_v02/troute/nhd_io.py
@@ -53,6 +53,7 @@ def read_custom_input(custom_input_file):
     restart_parameters = data.get("restart_parameters", {})
     output_parameters = data.get("output_parameters", {})
     run_parameters = data.get("run_parameters", {})
+    parity_parameters = data.get("parity_parameters", {})
     # TODO: add error trapping for potentially missing files
     return (
         supernetwork_parameters,
@@ -61,6 +62,7 @@ def read_custom_input(custom_input_file):
         restart_parameters,
         output_parameters,
         run_parameters,
+        parity_parameters,
     )
 
 
@@ -136,7 +138,7 @@ def read_qlat(path):
     return get_ql_from_csv(path)
 
 
-def get_ql_from_wrf_hydro_mf(qlat_files, index_col="station_id", value_col="q_lateral"):
+def get_ql_from_wrf_hydro_mf(qlat_files, index_col="station_id", value_col="q_lateral", filter_list=None):
     """
     qlat_files: globbed list of CHRTOUT files containing desired lateral inflows
     index_col: column/field in the CHRTOUT files with the segment/link id
--- a/src/python_framework_v02/troute/nhd_io.py
+++ b/src/python_framework_v02/troute/nhd_io.py
@@ -185,10 +185,17 @@ def get_ql_from_wrf_hydro_mf(qlat_files, index_col="station_id", value_col="q_la
     #       method should be better for all cases.
     if len(qlat_files) == 1:
         with xr.open_mfdataset(qlat_files, combine='nested', concat_dim='time') as ds:
-            df1 = ds[value_col].to_dataframe()
+            if not filter_list:
+                df1 = ds[value_col].to_dataframe()
+            else:
+                df1 = ds.sel({index_col:filter_list})[value_col].to_dataframe()
+
     else: # Assumes > 1
         with xr.open_mfdataset(qlat_files, combine='by_coords') as ds:
-            df1 = ds[value_col].to_dataframe()
+            if not filter_list:
+                df1 = ds[value_col].to_dataframe()
+            else:
+                df1 = ds.sel({index_col:filter_list})[value_col].to_dataframe()
 
     mod = df1.reset_index()
     ql = mod.pivot(index=index_col, columns="time", values=value_col)
@@ -213,7 +220,7 @@ def get_ql_from_wrf_hydro(qlat_files, index_col="station_id", value_col="q_later
 
     for filename in qlat_files:
         with xr.open_dataset(filename) as ds:
-            df1 = ds.to_dataframe()
+            df1 = ds[value_col].to_dataframe()
 
         li.append(df1)
 
--- a/src/python_framework_v02/troute/nhd_io.py
+++ b/src/python_framework_v02/troute/nhd_io.py
@@ -139,7 +139,7 @@ def read_qlat(path):
 
 
 def get_ql_from_wrf_hydro_mf(
-    qlat_files, index_col="station_id", value_col="q_lateral", filter_list=None
+    qlat_files, index_col="feature_id", value_col="q_lateral"
 ):
     """
     qlat_files: globbed list of CHRTOUT files containing desired lateral inflows
@@ -153,8 +153,9 @@ def get_ql_from_wrf_hydro_mf(
     `get_ql_from_csv`
 
 
-    Note:
-    For later needs, filtering may be accomplished with one of:
+    Note/Todo:
+    For later needs, filtering for specific features or times may
+    be accomplished with one of:
         ds.loc[{selectors}]
         ds.sel({selectors})
         ds.isel({selectors})
@@ -163,7 +164,7 @@ def get_ql_from_wrf_hydro_mf(
 
     For example:
     ```
-    (Pdb) ds.sel({"feature_id":[4186117],"time":ds.time.values[:2]})['q_lateral'].to_dataframe()
+    (Pdb) ds.sel({"feature_id":[4186117, 4186169],"time":ds.time.values[:2]})['q_lateral'].to_dataframe()
                                      latitude  longitude  q_lateral
     time                feature_id
     2018-01-01 13:00:00 4186117     41.233807 -75.413895   0.006496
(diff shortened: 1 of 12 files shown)


jameshalgren:
Tagging contributors:
@awlostowski-noaa 
@jhrehanoaa 
@groutr 
@jdhondia-noaa 
@kumdonoaa

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -53,6 +53,7 @@ def read_custom_input(custom_input_file):
     restart_parameters = data.get("restart_parameters", {})
     output_parameters = data.get("output_parameters", {})
     run_parameters = data.get("run_parameters", {})
+    parity_parameters = data.get("parity_parameters", {})
     # TODO: add error trapping for potentially missing files
     return (
```
groutr, line 58: You might think about wrapping these return parameters in a dataclass or namedtuple. It will help you keep things organized.

src/python_framework_v02/troute/nhd_network.py
```diff
@@ -419,3 +426,86 @@ def replace_waterbodies_connections(connections, waterbodies):
...
+            # Build dict object containing reachable nodes within max_depth from each source in new_sources
+            rv = {}
+            for h in new_sources:
+
+                reachable = set()
+                Q = deque([(h, 0)])
+                stop_depth = 1000000
+                while Q:
+
+                    x, y = Q.popleft()
```
groutr, line 470: Minor nitpick, more descriptive variable names than `x` and `y` would help make this logic clearer.
the first element of the tuple is the node and the second element represents its depth. I would suggest renaming y -> us_depth and x -> node.

src/python_framework_v02/troute/nhd_network.py
```diff
@@ -419,3 +426,86 @@ def replace_waterbodies_connections(connections, waterbodies):
...
+                    reachable.add(x)
+
+                    rx = rconn.get(x, ())
```
groutr, line 473: This should be an if-statement check. Specifically, we need to ensure that we only extend Q when there are nodes upstream.

groutr:
```python
if x in rconn:
    rx = rconn[x]
    if len(rx) > 1:
        depth += 1
    Q.extend(zip(rx, [us_depth]*len(rx)))
```

src/python_framework_v02/troute/nhd_network.py
groutr:
```python
node, depth = Q.popleft()
reachable.add(node)
...
```
This also removes the need to set `us_depth` inside the loop.
```python
if len(rx) > 1:
    depth += 1
```

src/python_routing_v02/fast_reach/mc_reach.pyx
```diff
@@ -348,7 +392,16 @@ cpdef object compute_network(
...
 
-    return np.asarray(data_idx, dtype=np.intp), np.asarray(flowveldepth, dtype='float32')
+
+    # delete the duplicate results that shouldn't be passed along
+    # The upstream keys have empty results because they are not part of any reaches
+    # so we need to delete the null values that return
+    if len(fill_index_mask) > 0:
+        data_idx_ma = [ix for i, ix in enumerate(data_idx) if i not in fill_index_mask]
+        flowveldepth_ma = [ix for i, ix in enumerate(flowveldepth) if i not in fill_index_mask]
+        return [np.asarray(data_idx_ma, dtype=np.intp), np.asarray(flowveldepth_ma, dtype='float32')]
```
groutr, line 402: Leave this return value as a tuple.

src/python_routing_v02/fast_reach/mc_reach.pyx
```diff
@@ -348,7 +392,16 @@ cpdef object compute_network(
...
+    else:
+        return [np.asarray(data_idx, dtype=np.intp), np.asarray(flowveldepth, dtype='float32')]
```
groutr, line 404: Remove the square brackets.

src/python_routing_v02/fast_reach/mc_reach.pyx
```diff
@@ -348,7 +392,16 @@ cpdef object compute_network(
 
             timestep += 1
 
-    return np.asarray(data_idx, dtype=np.intp), np.asarray(flowveldepth, dtype='float32')
+
+    # delete the duplicate results that shouldn't be passed along
+    # The upstream keys have empty results because they are not part of any reaches
+    # so we need to delete the null values that return
+    if len(fill_index_mask) > 0:
+        data_idx_ma = [ix for i, ix in enumerate(data_idx) if i not in fill_index_mask]
```
groutr, line 400: Here we are reading data out of a numpy array into a list and then back into a numpy array. I wonder if it would be faster to construct a mask and then mask data_idx.

groutr:
```python
fill_mask = np.ones_like(data_idx)
fill_mask[list(fill_index_mask)] = 0
return np.asarray(data_idx[fill_mask], dtype=np.intp), np.asarray(flowveldepth_ma[fill_mask], dtype='float32')
```

groutr: Looking at how fill_index_mask is constructed, we could construct the boolean mask directly I think.

jameshalgren:
ping @jmmcreight

awlostowski-noaa (dismissed):

src/python_routing_v02/build_tests.py
```diff
@@ -115,69 +113,102 @@ def build_test_parameters(test_name,
...
-        dt_wrf = (wrf_time[1] - wrf_time[0])
-        sim_duration = (wrf_time[-1] + dt_wrf) - wrf_time[0]
-        
-        dt = 300
-        dt_routing = pd.Timedelta(str(dt) + 'seconds')
-        
-        run_parameters["dt"] = dt
-        run_parameters["nts"] = round(sim_duration / dt_routing)
-        run_parameters["qts_subdivisions"] = dt_wrf/dt_routing
-
```
jameshalgren, line 139: Let's retain these with the other commented text.

src/python_routing_v02/fast_reach/mc_reach.pyx
jameshalgren: removed.

src/python_routing_v02/fast_reach/mc_reach.pyx
jameshalgren: removed.

src/python_framework_v02/troute/nhd_network.py
jameshalgren: Let's address this in an separate update to build_subnetworks.

src/python_framework_v02/troute/nhd_network.py
jameshalgren: Created issue, will address in PR focused on refactor of build_subnetworks function.

src/python_framework_v02/troute/nhd_io.py
jameshalgren: will address in #189

src/python_routing_v02/fast_reach/mc_reach.pyx
jameshalgren: Will respond with #205

src/python_routing_v02/build_tests.py
jameshalgren: Retained. (ping @awlostowski-noaa )

awlostowski-noaa (approved):

Issue #204, opened 2021-01-13
add if-statement check to Q.extend in build_subnetworks


jameshalgren:
This should be an if-statement check. Specifically, we need to ensure that we only extend Q when there are nodes upstream.

```python
if x in rconn:
    rx = rconn[x]
    if len(rx) > 1 and len(reachable) < min_size:  # we are at a junction
        Q.extend(zip(rx, [depth + 1] * len(rx)))
    else:
        Q.extend(zip(rx, [depth] * len(rx)))
```
Is tracking `stop_depth` really necessary?

_Originally posted by @groutr in https://github.com/NOAA-OWP/t-route/pull/188#discussion_r555852480_

awlostowski-noaa:
noted.

Issue #203, opened 2021-01-13
More descriptive variable names in build_subnetworks


jameshalgren:
Minor nitpick, more descriptive variable names than `x` and `y` would help make this logic clearer.
the first element of the tuple is the node and the second element represents its depth. I would suggest renaming y -> depth and x -> node.

```python
node, depth = Q.popleft()
reachable.add(node)
...
```
This also removes the need to set `us_depth` inside the loop.
```python
if len(rx) > 1:
    depth += 1
```

_Originally posted by @groutr in https://github.com/NOAA-OWP/t-route/pull/188#r555856866_

PR #218, opened 2021-01-14
[WIP -- Hold] Upgrade performance of subnetwork construction algorithm

awlostowski-noaa:
1. An if-statement was added around `Q.extend`, such that the deque is only extended if there are upstream connections. 
2. Several variables were renames to be more descriptive. For example, `x` and `y` were renamed to `node` and `depth`, respectively. 

Additionally, @groutr was wondering it it was necessary to track `stop_depth`. I think it is necessary, otherwise the subnetwork grouping routine continues until all nodes in the subnetwork have been reached. Besides, it less "tracking" and more a single update once the minimum subnetwork size has been exceeded. These are my thoughts, but I am a Python tyro. 

Addressing #204 and #203

### DIFF
--- a/src/python_framework_v02/troute/nhd_network.py
+++ b/src/python_framework_v02/troute/nhd_network.py
@@ -428,6 +428,9 @@ def replace_waterbodies_connections(connections, waterbodies):
     return new_conn
 
 
+# TO DO: 
+# - add if statement check to Q.extend
+# more descriptive variable names
 def build_subnetworks(connections, rconn, min_size, sources=None):
     """
     Construct subnetworks using a truncated breadth-first-search
--- a/src/python_framework_v02/troute/nhd_network.py
+++ b/src/python_framework_v02/troute/nhd_network.py
@@ -455,7 +455,6 @@ def build_subnetworks(connections, rconn, min_size, sources=None):
     for net in sources:
 
         # subnetwork creation using a breadth first search restricted by maximum allowable depth
-        # new_sources_list = [net]
         new_sources = set([net])
         subnetworks = {}
         group_order = 0
@@ -468,22 +467,29 @@ def build_subnetworks(connections, rconn, min_size, sources=None):
                 reachable = set()
                 Q = deque([(h, 0)])
                 stop_depth = 1000000
+                
                 while Q:
-
-                    x, y = Q.popleft()
-                    reachable.add(x)
-
-                    rx = rconn.get(x, ())
-                    if len(rx) > 1:
-                        us_depth = y + 1
-                    else:
-                        us_depth = y
-
-                    if len(reachable) > min_size:
-                        stop_depth = y
-
-                    if us_depth <= stop_depth:
-                        Q.extend(zip(rx, [us_depth] * len(rx)))
+                    
+                    # pop node and it's topological depth from Q, add node to reachable list
+                    node, depth = Q.popleft()
+                    reachable.add(node)
+                    
+                    # get upstream connections of current node
+                    us_nodes = rconn.get(node, ())
+                    
+                    # extend Q if there are nodes upstream
+                    if len(us_nodes) > 0:  
+                        # upstream depth increases if a junction is traversed
+                        if len(us_nodes) > 1:
+                            us_depth = depth + 1
+                        else:
+                            us_depth = depth
+                        # enforce minimum subnetwork size requirement
+                        if len(reachable) > min_size:
+                            stop_depth = depth
+                        # extend Q with upstream nodes and their topological depths
+                        if us_depth <= stop_depth:
+                            Q.extend(zip(us_nodes, [us_depth] * len(us_nodes)))
 
                 # reachable: a list of reachable segments within max_depth from source node h
                 rv[h] = reachable


src/python_framework_v02/troute/nhd_network.py
```diff
@@ -465,22 +467,29 @@ def build_subnetworks(connections, rconn, min_size, sources=None):
...
+                    # extend Q if there are nodes upstream
+                    if len(us_nodes) > 0:  
+                        # upstream depth increases if a junction is traversed
+                        if len(us_nodes) > 1:
+                            us_depth = depth + 1
+                        else:
+                            us_depth = depth
+                        # enforce minimum subnetwork size requirement
+                        if len(reachable) > min_size:
+                            stop_depth = depth
```
awlostowski-noaa, line 489: perhaps superfluous

groutr:
If I understood your original correctly, `stop_depth` was necessary because `Q` wasn't being managed correctly. If `rx` was empty, we would still extend `Q` with an empty pair if we hadn't reached our threshold. Thus `stop_depth` was added remedy that fact that `Q` wasn't reflecting the true amount of work to do. Now if `Q` is being managed correctly, when `Q` is _really_ empty (which is what the while condition checks), we've reached our stop depth.

awlostowski-noaa:
The subnetwork build stops when either
1. there are no more upstream junctions to step to, or
2. the minimum subnetwork size (number of segments) has been exceeded.

`stop_depth` is only updated when `min_size` is exceeded:
```python
# enforce minimum subnetwork size requirement
if len(reachable) > min_size:
    stop_depth = depth
```
This simple statement updates stop_depth to the current subnetwork depth, such that Q is not extended. 

While this might be a bit "hacky" it works and seems to require minimal effort. I've tried ways to work around needing `stop_depth`, but have been unsuccessful.

awlostowski-noaa:
ping @jameshalgren

groutr:
@awlostowski-noaa I think I misunderstood the intent of your algorithm. As I understand things now, your algorithm attempts to return all the nodes up to a specific order (note that I'm using the term order instead of depth!). 

Given the following full binary tree (a special case where every node is a junction)
```
{0: [1, 2], 1: [3, 4], 2: [5, 6], 3: [7, 8], 4: [9, 10], 5: [11, 12], 6: [13, 14], 7: [], 8: [], 9: [], 10: [], 11: [], 12: [], 13: [], 14: []}
```
If we want to have a min size 3, we would expect `{0, 1, 2}` and a min size of 4 should produce `{0, 1, 2, 3, 4, 5, 6}`
I think there is a bug in both of our candidates for a traversal function.

I decided to factor some your code out into another function for better clarity. Consider the following functions:
```python
def adam_bfs(rconn, h, min_size):
    reachable = set()
    Q = deque([(h, 0)])
    stop_depth = 1000000

    while Q:

        # pop node and it's topological depth from Q, add node to reachable list
        node, depth = Q.popleft()
        reachable.add(node)

        # get upstream connections of current node
        us_nodes = rconn.get(node, ())

        # extend Q if there are nodes upstream
        if len(us_nodes) > 0:  
            # upstream depth increases if a junction is traversed
            if len(us_nodes) > 1:
                us_depth = depth + 1
            else:
                us_depth = depth
            # enforce minimum subnetwork size requirement
            if len(reachable) > min_size:
                stop_depth = depth
            # extend Q with upstream nodes and their topological depths
            if us_depth <= stop_depth:
                Q.extend(zip(us_nodes, [us_depth] * len(us_nodes)))
    return reachable

def threshold_limited_bfs(connections, source, size=None):
    """Visit all nodes until size has been reached. Then finish visiting the current order and return"""
    if size is None:
        size = float('inf')
    
    visited = set()
    Q = deque([(source, 0)])
    curr_order = 0
    while Q:
        node, order = Q.popleft()
        
        # Have we crossed a junction and we have visited enough nodes?
        if order > curr_order and len(visited) >= size:
            break
        else:
            curr_order = order
        
        # We visit node *after* we've checked whether to continue.
        visited.add(node)
        
        if node in connections:
            children = connections[node]
            if len(children) > 1:
                Q.extend(zip(children, repeat(order+1)))
            else:
                Q.extend(zip(children, repeat(order)))
    return visited
```

I believe that my previous suggestion and your current implementation are incorrect. I submit `threshold_limited_bfs` as a correct implementation.

In this example observe how we accidentally visit too many nodes with `adam_bfs`. Nodes 7 and 8 should not have been returned because they cross into the next order.
```python
>>> adam_bfs(tree, 0, 4)
{0, 1, 2, 3, 4, 5, 6, 7, 8}
>>> threshold_limited_bfs(tree, 0, size=4)
{0, 1, 2, 3, 4, 5, 6}
```
In this test case, `adam_bfs` actually goes one level too far. Nodes 3, 4, 5, and 6 actually belong to the next order.
```python
>>> adam_bfs(tree, 0, 3)
{0, 1, 2, 3, 4, 5, 6}
>>> threshold_limited_bfs(tree, 0, size=3)
{0, 1, 2}
```

jameshalgren:
Ways to consider to finish this PR: 
1 Stop and close it and make new issues/PRs for this list
1 Figure out how to implement the algorithm by @groutr, otherwise preserving the function of `build_subnetworks`
1 Re-do `build_subnetworks` to eliminate some of the post-processing steps now required
1 Add arbolate sum to the dfs search and make a new, more efficient `build_subnetworks_optimized` function.
 
I would vote for __4__ because the real challenge we face in terms of computation time is in the iterations, and if we can improve the iteration time with a more effective subnetwork set, that would be a potentially huge beneft. 

Keywords I've found include:
- arbolate sum (seems to be exclusively USGS - refers to the length of tree graph above and up to a node)
- tree graph (what we have -- "In graph theory, a tree is an undirected graph in which any two vertices are connected by exactly one path, or equivalently a connected acyclic undirected graph.)
- forest -- a collection of trees. I've been calling each tree a "network" and the forest a "supernetwork" 

There are several references that have been helping me ponder. None seems to have the silver bullet to achieving even distribution of subnetworks for optimum calculation speed. The first one seems to head down the right path. 
https://core.ac.uk/download/pdf/147691424.pdf
https://stackoverflow.com/questions/2225865/strategy-to-implement-tree-traversing-algorithm-in-parallel
http://www.shodor.org/media/content/petascale/materials/UPModules/Binary_Tree_Traversal/Binary_Tree_Traversal_Module_Document.pdf
https://ieeexplore.ieee.org/document/147799
https://medium.com/nothingaholic/depth-first-search-vs-breadth-first-search-in-python-81521caa8f44

jameshalgren:
Documenting some effort to integrate the modified algorithm:  
For the Pocono1 test case(`python compute_nhd_routing_SingleSeg_v02.py -t -v --debuglevel 1 --parallel by-subnetwork-jit-clustered`), the rconn is as follows: 
```
(Pdb) pp rconn
defaultdict(None,
            {4153018: [],
             4153786: [],
             4185517: [],
             4185527: [],
             4185583: [4185527],
             4185587: [4185517],
             4185607: [4186255],
             4185613: [],
             4185649: [],
             4185661: [4186257],
             4185663: [4186499],
             4185665: [4185661, 4185663],
             4185667: [4185613],
             4185693: [4186503],
             4186251: [4185587],
             4186253: [4185583],
             4186255: [4186251, 4186253],
             4186257: [4185607],
             4186287: [4185665],
             4186297: [4185693],
             4186305: [4186287, 4186297],
             4186307: [4185667],
             4186309: [4186305, 4186307],
             4186499: [],
             4186503: [932040249],
             932040249: [932040251],
             932040251: [932040253],
             932040253: [4185649]})

```
... and the levels/reach orders for that collection may be outlined as follows.  I am testing now to make sure the bfs algorithm does not split the long sequences ... 
```
Levels:
0   1   2   3   4
4186309: [4186305, 4186307],
├── 4186305: [4186287, 4186297],
|   ├── 4186297: [4185693],
|   |   ├4185693: [4186503],
|   |   ├4186503: [932040249],
|   |   ├932040249: [932040251],
|   |   ├932040251: [932040253],
|   |   ├932040253: [4185649]})
|   |   └4185649: [],
|   └── 4186287: [4185665],
|       └4185665: [4185661, 4185663],
|        ├─ 4185663: [4186499],
|        |  └4186499: [],
|        └─ 4185661: [4186257],
|           ├4186257: [4185607],
|           ├4185607: [4186255],
|           └4186255: [4186251, 4186253],
|            ├─ 4186251: [4185587],
|            |  ├4185587: [4185517],
|            |  └4185517: [],
|            └─ 4186253: [4185583],
|               ├4185583: [4185527],
|               └4185527: [],
└── 4186307: [4185667],
    ├4185667: [4185613],
    └4185613: [],
```

jameshalgren:
@awlostowski-noaa The current work on network splitting for diffusive is related to this PR. We might try to find a way to further modularize the network splitting so that this and other work can all be combined.

PR #187, opened 2020-12-18
Incorporate direct insertion of USGS timeslice data

jhrehanoaa:
New WIP branch that implements the globbed usgs timeslice data into a dataframe. Still need to trim the column names from the strange b' added and we need to look at the format of the df and if we need to change it around. Currently, it displays discharge and id by timeslice. I have a print statement in it to show the df as is.

PR #258, opened 2021-02-26
Add data assimilation using USGS timeslices

jameshalgren:
supersedes #187 
On approval -- *squash and merge!* (there are some unusual tree relics here that we do not need to preserve.)

This PR creates basic data assimilation capability in v02 based on the usgs 'Nudging' pattern in the current NWM. 

A follow-on PR will provide a test case for this capability. 

Test cases that verify continued operation of existing functionality are:
- [x] python compute_nhd_routing_SingleSeg_v02.py -f ../../test/input/yaml/CustomInput.yaml
- [x] python compute_nhd_routing_SingleSeg_v02.py -t -n Mainstems_CONUS -v --debuglevel 1 --parallel by-subnetwork-jit-clustered --subnet-size 100 --nts 288 --cpu-pool 6
- [x] python compute_nhd_routing_SingleSeg_v02.py -t -n Mainstems_CONUS -v --debuglevel 1 --parallel by-network --nts 288 --cpu-pool 6
- [x] python compute_nhd_routing_SingleSeg_v02.py --test pocono1 -t -v --debuglevel 1
- [x] python compute_nhd_routing_SingleSeg_v02.py -f ../../test/input/yaml/Florence_Benchmark.yaml


jhrehanoaa (approved):

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -189,12 +193,20 @@ def get_ql_from_wrf_hydro_mf(qlat_files, index_col="feature_id", value_col="q_la
...
-            columns=ds.time.values,
-            # dtype=float,
-        )
+        try:
+            ql = pd.DataFrame(
+                ds[value_col].values.T,
+                index=ds[index_col].values[0],
+                columns=ds.time.values,
+                # dtype=float,
+            )
```
jameshalgren, line 202: Dig deeper into this difference in the qlateral files -- is there an intentional difference between the output from the RnR method and the files we are receiving from the later model runs?

The except statement works with the more current files. The try statement works with the RnR output -- it fails quickly on a type mismatch, so it is not causing a significant performance issue.

PR #264, opened 2021-03-04
qlat filtering by mask file

jhrehanoaa:
same as q0 filtering by mask file but with qlats


src/python_framework_v02/troute/nhd_network_utilities_v02.py
```diff
@@ -550,6 +550,12 @@ def build_qlateral_array(forcing_parameters, connections_keys, nts, qts_subdivis
     if len(qlat_df.columns) > max_col:
         qlat_df.drop(qlat_df.columns[max_col:], axis=1, inplace=True)
 
+    mask_file_path = supernetwork_parameters.get("mask_file_path", None)
+    if mask_file_path:
+        mask_file_path = pd.read_csv(mask_file_path)
+        mask_file_path = mask_file_path.iloc[:, 0].tolist()
```
jameshalgren: Could we test `mask_file_path.index.values()`?

jameshalgren (approved):
Great. Thank you.

PR #266, opened 2021-03-10
Self loop capability

jhrehanoaa:
Continuing work on self looping from restart files. Organizational changes still need to be made and some bug fixes.

### DIFF
--- a/src/python_routing_v02/compute_nhd_routing_SingleSeg_v02.py
+++ b/src/python_routing_v02/compute_nhd_routing_SingleSeg_v02.py
@@ -824,16 +824,13 @@ def main():
         start_time = time.time()
     if verbose:
         print("setting channel initial states ...")
-    import pdb; pdb.set_trace()
+
     if ts_iterator == 0:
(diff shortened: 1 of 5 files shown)


src/python_routing_v02/compute_nhd_routing_SingleSeg_v02.py
```diff
@@ -608,6 +609,7 @@ def compute_nhd_routing_v02(
 
     else:  # Execute in serial
         results = []
+
         for twi, (tw, reach_list) in enumerate(reaches_bytw.items(), 1):
             segs = list(chain.from_iterable(reach_list))
             param_df_sub = param_df.loc[
```
jhrehanoaa, line 615: I've run into an issue here regarding the segs list. When a run is performed and an output csv file is created from the results the index changes. At the end of a run we output to csv using restart_flows as the current index, but this doesn't match the original restart file. When running a second iteration we run into an issue here in reach_list from reaches_bytw. What would be the proper way to modify this to either recreate the original restart index values of the original restart file or modify the reach_list/reaches_bytw to reflect the current iteration?

PR #268, opened 2021-03-10
Speedup revision to nhd_network_augment

awlostowski-noaa:
This pull request contains many changes to `nhd_network_augment.py` that are focused on expediting the execution time, such that  it will run on the CONUS domain in a reasonable amount of time. Most of the changes fall into one of two categories:
1) Pulling pandas indexing operations out of loops. Turns out, indexing a pandas DataFrame is computationally burdensome when it is done over and over again many thousands of times in a loop. By moving any and all pandas DataFrame indexing out of loops and using pandas to work with bulk information edits, much speed up is required. 
2) Using numpy arrays rather than pandas DataFrames for in-loop indexing

### DIFF
--- a/src/python_framework_v02/troute/nhd_network_augment.py
+++ b/src/python_framework_v02/troute/nhd_network_augment.py
@@ -405,47 +405,24 @@ def merge_parameters(to_merge):
     Returns:
         replace (DataFrame): weighted average
     """
-
-#     data_replace = to_merge.tail(1)
-#     data_replace._is_copy = None
-    
-    data_replace = pd.DataFrame().reindex_like(to_merge.tail(1))
-
-#     idx = to_merge.tail(1).index
     idx = to_merge.index[-1]
-
-    data_replace.loc[idx,"to"] = to_merge.loc[idx,"to"]
-    data_replace.loc[idx,"Length"] = to_merge.Length.sum()
-    data_replace.loc[idx,"n"] = len_weighted_av(to_merge, "n", "Length")
-    data_replace.loc[idx,"nCC"] = len_weighted_av(to_merge, "nCC", "Length")
-    data_replace.loc[idx,"So"] = len_weighted_av(to_merge, "So", "Length")
-    data_replace.loc[idx,"BtmWdth"] = len_weighted_av(to_merge, "BtmWdth", "Length")
-    data_replace.loc[idx,"TopWdth"] = len_weighted_av(to_merge, "TopWdth", "Length")
-    data_replace.loc[idx,"TopWdthCC"] = len_weighted_av(
+    data_replace = to_merge.loc[idx].to_numpy()
+    
+    data_replace[1] = to_merge.Length.sum()
+    data_replace[2] = len_weighted_av(to_merge, "n", "Length")
+    data_replace[3] = len_weighted_av(to_merge, "nCC", "Length")
+    data_replace[4] = len_weighted_av(to_merge, "So", "Length")
+    data_replace[5] = len_weighted_av(to_merge, "BtmWdth", "Length")
+    data_replace[6] = len_weighted_av(to_merge, "TopWdth", "Length")
+    data_replace[7] = len_weighted_av(
         to_merge, "TopWdthCC", "Length"
     )
-    data_replace.loc[idx,"NHDWaterbodyComID"] = to_merge.loc[idx,"NHDWaterbodyComID"]
-    data_replace.loc[idx,"MusK"] = to_merge.loc[idx,"MusK"]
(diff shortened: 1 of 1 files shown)


src/python_framework_v02/troute/nhd_network_augment.py
```diff
@@ -159,49 +167,138 @@ def build_reaches(rconn):
...
+    Construct a list of headwater reaches
+    Args:
+        subreaches (dict): reaches (linearly connected segments) connected to each network tailwater (keys)
+        hws (list): headwater segments
+    Returns:
+        hw_reaches (list): lists of headwater reaches 
+    """
+    
+    hw_reaches = []
+    for sublists in list(subreaches.values()):
```
groutr, line 185: Don't create a list object here.

src/python_framework_v02/troute/nhd_network_augment.py
```diff
@@ -159,49 +167,138 @@ def build_reaches(rconn):
...
+        for rch in sublists:
+            for val in rch:
```
groutr, line 187: You are essentially doing a set intersection here. It might be faster to use sets than do linear searches.
```
...
for rch in sublists:
    rch = set(rch)
    hw_reaches.extend(rch & hws)
```

groutr: Creating the set is O(n) (and happens once), but lookups are O(1) so looking up m values will be O(n+m).
Iterating thru rch is O(n), but for each value we are executing a linear search of m items (hws) => O(n*m)

groutr: Actually, make hws the set.
```
def find_headwater_reaches(...):
    hws = set(hws)
    for sublists in subreaches.values():
        for rch in sublists:
            hw_reaches.extend(hws & rch)
    return hw_reaches
```

src/python_framework_v02/troute/nhd_network_augment.py
```diff
@@ -159,49 +167,138 @@ def build_reaches(rconn):
...
+        threshold (int): threshold length that defines short
+    Returns:
+        short_hw_reaches (list): lists of SHORT headwater reaches 
+    """
+    
+    short_hw_reaches = []
+    for rch in hw_reaches:
+
+        # find row locations of reach segments
+        a = np.searchsorted(idx_array, rch)
```
groutr, line 211: Do we have a guarantee that idx_array is sorted? If idx_array not sorted, np.searchsorted will fail.

src/python_framework_v02/troute/nhd_network_augment.py
awlostowski-noaa: hmm. This suggestion seems to make `hw_reaches` just a list of headwater segments. Instead, we'd want `hw_reaches` to be a list of lists, where each list is a combination of connected segments comprising a reach, or a set of tuples, where each tuple is a collection of connected segments comprising a reach. 

Also, `rch` is a list so the `&` operation fails, unless we convert to a set.

src/python_framework_v02/troute/nhd_network_augment.py
awlostowski-noaa: Yes. `idx_array` is extracted from the network data DataFrame constructed and returned by `def get_network_data`.  In this function, the DataFrame is sorted by index - ensuring that `idx_array` will also be sorted.

src/python_framework_v02/troute/nhd_network_augment.py
groutr: Use the set.intersection method. It accepts list. Change extend to append.
We know that headwaters can only be the first element of the reach list. Can we just check that element rather than looping thru the reach?

src/python_framework_v02/troute/nhd_network_augment.py
```diff
@@ -764,28 +781,88 @@ def main():
...
+        
+        tws_N1 = count_tailwaters(N1, network_data)
+        tws_N2 = count_tailwaters(N2, network_data)
+        return(tws_N1 == tws_N2)
+    
+    check = connection_check(data_snapped, data_merged, network_data)
+    if check:
+        print("Passed connection test")
+    else:
+        print("Error! - The merging process has created additional tailwaters (i.e. it broke the network)")
```
jameshalgren, line 819: Good idea to check this!

jameshalgren:
Broad comment here is that a number of the options which were used in getting to the endpoint are no longer needed or at least are no longer functional. Given the low priority nature of incorporating this in the repo, we can take our time clearing those things out and merge this after that.

Issue #251, opened 2021-02-22
Remove redundant arguments


jameshalgren:
We need only pass `run_parameters` -- the `return_courant`, `nts`, and `dt` are all extractable from there as subelements. 

run_parameters.get["return_courant", False"]
run_parameters["nts"],
run_parameters["dt"],

https://github.com/NOAA-OWP/t-route/blob/afd4ee3502b3c214dc6c4b81b62892eefc3e471a/src/python_routing_v02/build_tests.py#L160

jhrehanoaa:
closed by #269

PR #269, opened 2021-03-12
removed redundant arguments nts dt

jhrehanoaa:
fixes issue #251

### DIFF
--- a/src/python_routing_v02/build_tests.py
+++ b/src/python_routing_v02/build_tests.py
@@ -53,10 +53,7 @@ def build_test_parameters(
 
         # Simulation domain RouteLink file
         routelink_file = pathlib.Path(
-            NWM_test_path,
-            "primary_domain",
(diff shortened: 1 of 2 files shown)


jameshalgren (approved):
All good. Thank you.

src/python_routing_v02/build_tests.py
```diff
@@ -160,13 +157,15 @@ def build_test_parameters(
     )
 
 
-def parity_check(parity_parameters, run_parameters, nts, dt, results):
-    
+def parity_check(parity_parameters, run_parameters, results):
+    nts = run_parameters["nts"]
+    dt = run_parameters["dt"]
+
```
jameshalgren, line 163: These three lines (the `parity_check` function and the two variables below that) seem to have been dropped from the merge or were stomped on by a later commit.
@jhrehanoaa

PR #273, opened 2021-03-18
Add data assimilation using usgs timeslices

jhrehanoaa:
https://github.com/NOAA-OWP/t-route/pull/258#discussion_r584912652 - From this discussion. 

RnR file can run with DA, but no matching columns in the parity check. I am wondering what the locations are for the timeslice files for this RnR? RnR had 'feature_id' instead of 'station_id' as well. Currently, this runs, but the parity test is comparing two totally different things essentially.

### DIFF
--- a/src/python_framework_v02/troute/nhd_io.py
+++ b/src/python_framework_v02/troute/nhd_io.py
@@ -85,6 +85,7 @@ def read_waterbody_df(waterbody_parameters, waterbodies_values, wbtype="level_po
     General waterbody dataframe reader. At present, only level-pool
     capability exists.
     """
+
     if wbtype == "level_pool":
         wb_params = waterbody_parameters[wbtype]
         return read_level_pool_waterbody_df(
@@ -141,7 +142,7 @@ def read_qlat(path):
     return get_ql_from_csv(path)
 
 
-def get_ql_from_wrf_hydro_mf(qlat_files, index_col="feature_id", value_col="q_lateral"):
+def get_ql_from_wrf_hydro_mf(qlat_files, index_col, value_col="q_lateral"):
     """
     qlat_files: globbed list of CHRTOUT files containing desired lateral inflows
     index_col: column/field in the CHRTOUT files with the segment/link id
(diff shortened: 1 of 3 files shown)


src/python_framework_v02/troute/nhd_io.py
```diff
@@ -180,8 +181,8 @@ def get_ql_from_wrf_hydro_mf(qlat_files, index_col="feature_id", value_col="q_la
     2018-01-01 13:00:00 4186117     41.233807 -75.413895   0.006496
     ```
     """
-    filter_list = None
 
+    filter_list = None
```
jameshalgren, line 185: I don't think we are using this parameter anywhere, so we can probably remove it.

jameshalgren:
@jhrehanoaa 
Has this become a PR for integrating a test case for DA?

If so, could we rebase and start working on it from that angle?

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -322,27 +320,23 @@ def get_usgs_from_time_slices_csv(
         usgs_df.insert(i + 11, temp_name8, np.nan)
 
     usgs_df = usgs_df.interpolate(method="linear", axis=1)
-    usgs_df.drop(usgs_df[usgs_df.iloc[:,0] == -999999.000000].index , inplace=True)
+    usgs_df.drop(usgs_df[usgs_df.iloc[:, 0] == -999999.000000].index, inplace=True)
```
groutr, line 323: If we are computing a mask, a call to drop is unnecessary
Assuming that first column of usgs_df is labeled "ColumnA"
```
usgs_df = usgs_df.loc[usgs_df["ColumnA"] == -999999]
```

src/python_routing_v02/compute_nhd_routing_SingleSeg_v02.py
```diff
@@ -1044,18 +1044,19 @@ def main():
 
     if break_network_at_waterbodies:
         # Read waterbody parameters
-        waterbodies_df = nhd_io.read_waterbody_df(
-            waterbody_parameters, {"level_pool": wbodies.values()}
-        )
+        if wbodies != {}:
```
groutr, line 1047: `if wbodies:`

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -384,7 +378,7 @@ def get_usgs_from_time_slices_folder(
...
```
groutr, line 381: See above comment. The call to .drop can be avoided here also.

jameshalgren:
@jhrehanoaa -- I think this is a merge relic and am closing. Please reopen if this is still a relevant PR.

Issue #249, opened 2021-02-19
Use optimal 2-d array indexing in mc_reach


jameshalgren:
Several corrections were made in response to the original comment. A separate, simple PR should be used to fix this last holdout (and any others that might have crept in) in `mc_reach.pyx`: 
```212             flowveldepth[fill_index][idx] = val```
to be this: 
``` 212             flowveldepth[fill_index, idx] = val```

> Try using `buf_view[i, j]` instead of `buf_view[i][j]` for all of these memoryview accesses. If my memory is correct, `buf_view[i][j]` will create an intermediate object (`buf_view[i]` evaluates, then `tmp_obj[j]`).  Maybe cython is smart enough to optimize that away, but I doubt it.

_Originally posted by @groutr in https://github.com/NOAA-OWP/t-route/pull/224#discussion_r565835333_

JurgenZach-NOAA:
New array indexing has long been adopted in mc_reach.

PR #274, opened 2021-03-18
Use comma [i,j] index instead of [i][j]

jhrehanoaa:
change `flowveldepth[fill_index][idx] = val` to  `flowveldepth[fill_index, idx] = val`

#249 

Just made the tiny changed listed in the above comment.


jameshalgren (approved):
Thanks... simple and clean!

PR #279, opened 2021-04-12
Fix compile errors.

groutr:
This resolves the issues with missing symbols on my machine as well as
the troubles with the netcdf library.

The root of the issue was the hardcoding of gfortran in the makefiles. The netcdf library issues were resolved by querying netcdf library for the include paths and linking options.

tag: @hellkite500 @jameshalgren

### DIFF
--- a/src/python_routing_v02/compiler.sh
+++ b/src/python_routing_v02/compiler.sh
@@ -11,6 +11,12 @@ build_reservoir_kernel=true
 build_framework=true
 build_routing=true
 
+if [ -z "$F90" ]
+then
+    export F90="gfortran"
+    echo "using F90='gfortran'"
+fi
+
 #if you have custom static library paths, uncomment below and export them
 #export LIBRARY_PATH=<paths>:$LIBRARY_PATH
 #if you have custom dynamic library paths, uncomment below and export them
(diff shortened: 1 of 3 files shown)


jameshalgren (commented):
What would it look like to add a simple test to check if $F90 is empty, either in the makefile or in the compiler.sh script?

jameshalgren:
Tested on gfortran -- works. 

Compiles but produces the following error with intel compiler stack on Cheyenne (not a reason not to merge the PR.)

```
python compute_nhd_routing_SingleSeg_v02.py -f ../../test/input/yaml/CustomInput.yaml
Traceback (most recent call last):
  File "compute_nhd_routing_SingleSeg_v02.py", line 297, in <module>
    import fast_reach
ImportError: /glade/u/home/user/t-route/src/python_routing_v02/fast_reach.cpython-37m-x86_64-linux-gnu.so: undefined symbol: __intel_sse2_strlen
```

src/python_routing_v02/compiler.sh
```diff
@@ -10,6 +10,12 @@ build_reservoir_kernel=true
 build_framework=true
 build_routing=true
 
+if [ -z "$F90" ]
+then
+    export F90="gfortran"
+    echo "using F90='gfortran'"
+fi
+
```
jameshalgren, line 18: @groutr or @hellkite500 
Will you comment on the palatability of this insertion to catch the case where no F90 environment variable is in place?

Thanks...

hellkite500: This is really just a dev utility, I don't see too much harm imposing a default.  Worst case, gfortran doesn't exist and a make error occurs.

jameshalgren (approved):

jameshalgren (approved):
Rebased. Approved.

PR #299, opened 2021-04-22
C implementation of Muskingum Cunge

groutr:
Port the Muskingum Cunge computation from Cython to pure C.
The primary goal was a cleaner version of Muskingum Cunge that could be free of the Cython boilerplate code, and thus be more readily optimized by a C compiler. This implementation uses double precision floats instead of single precision.

Compiling with `-ffast-math` give a small, but noticeable performance improvement without any apparent loss of accuracy.

### DIFF
--- a/src/fortran_routing/diffusive_v02frwk/makefile
+++ b/src/fortran_routing/diffusive_v02frwk/makefile
@@ -1,5 +1,3 @@
-# compiler
-FC := gfortran
 
 # compile flags
 FCFLAGS = -g -c -O2 -fPIC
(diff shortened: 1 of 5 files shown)


jameshalgren:
Confirming -- @groutr -- this PR appears to make the introduction of these functions as capability but does not open any direct doors for their use in the existing workflow, is that correct?

groutr:
@jameshalgren I'm still moving bits an pieces over from the work on my previous branch. I appear to have lost track of some of my commits.

jameshalgren:
@JoshCu, @TrupeshKumarPatel 
Possibly a diversion, but also possibly worth resurrecting.

PR #300, opened 2021-04-23
Minor makefile standardization

jameshalgren:
These minor updates standardize (or attempt to standardize) the implicit variables as used in the various fortran makefiles. 

See https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html

### DIFF
--- a/src/fortran_routing/Reservoir_Binding/Reservoirs/makefile
+++ b/src/fortran_routing/Reservoir_Binding/Reservoirs/makefile
@@ -3,7 +3,7 @@ NETCDFLIB := $(shell nc-config --flibs)
 FC = $(F90)
 FFLAGS += -g -w -c -ffree-form -ffree-line-length-none -fconvert=big-endian -frecord-marker=4 $(NETCDFINC)
 LDFLAGS += $(NETCDFLIB) -static-libgfortran -static-libgcc -nodefaultlibs -lgfortran -lgcc -fPIC
-ARFLAGS += rc
+ARFLAGS := rc
 
 VPATH = ./Level_Pool ./Persistence_Level_Pool_Hybrid ./RFC_Forecasts
 
--- a/src/fortran_routing/Reservoir_Binding/Reservoirs/makefile
+++ b/src/fortran_routing/Reservoir_Binding/Reservoirs/makefile
@@ -39,10 +39,10 @@ bind_rfc.a: module_reservoir.o module_hydro_stop.o module_reservoir_utilities.o 
 	$(AR) $(ARFLAGS) $@ $^
 
 %.o: %.F
-	$(FC) $(FFLAGS) $(FFLAGS) $(LDFLAGS) -c -o $@ $<
+	$(FC) $(FFLAGS) $(LDFLAGS) -c -o $@ $<
 
 %.o: %.f90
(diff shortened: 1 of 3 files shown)


src/fortran_routing/Reservoir_Binding/Reservoirs/makefile
```diff
@@ -1,8 +1,8 @@
 NETCDFINC := $(shell nc-config --fflags)
 NETCDFLIB := $(shell nc-config --flibs)
-COMPILER90 = $(F90)
-F90FLAGS := -g -w -c -ffree-form -ffree-line-length-none -fconvert=big-endian -frecord-marker=4 $(NETCDFINC)
-F90LFLAGS := $(NETCDFLIB) -static-libgfortran -static-libgcc -nodefaultlibs -lgfortran -lgcc -fPIC
+FC = $(F90)
+FFLAGS += -g -w -c -ffree-form -ffree-line-length-none -fconvert=big-endian -frecord-marker=4 $(NETCDFINC)
```
hellkite500, line 4: Why += here?  `:=` is an immediate assignment, which takes place in the first pass parse of the makefile and all references are replaced with those contents.  `+=` is likely to result in a recursive, deffered variable that has to be re-evaluated for each refernce.

jameshalgren: Good question. The only answer that I can give is this, where I found the suggestion to use `+=`: 
https://www.gnu.org/software/make/manual/html_node/Appending.html

I'm open to either, especially given that it appears that the `+=` is mostly for specialized cases. But it does seem to be the preferred/recommended method in that article.

src/fortran_routing/Reservoir_Binding/Reservoirs/makefile
```diff
@@ -1,9 +1,9 @@
 NETCDFINC := $(shell nc-config --fflags)
 NETCDFLIB := $(shell nc-config --flibs)
-COMPILER90 = $(F90)
-F90FLAGS := -g -w -c -ffree-form -ffree-line-length-none -fconvert=big-endian -frecord-marker=4 $(NETCDFINC)
-F90LFLAGS := $(NETCDFLIB) -static-libgfortran -static-libgcc -nodefaultlibs -lgfortran -lgcc -fPIC
```
jameshalgren, line 5: regarding the linking here, do we need to use both `-static-libgfortran` and `-lgfortran`? Likewise for `-static-libgcc` and `-lgcc`...?

jameshalgren:
Merged -- let's move any further comment to the c_mc PR. #299.

src/fortran_routing/Reservoir_Binding/Reservoirs/makefile
jdmattern-noaa: Would have to defer to @hellkite500

src/fortran_routing/Reservoir_Binding/Reservoirs/makefile
jameshalgren: Another reference ... again, I'm having a hard time teasing a solid recommendation one way or the other from this, so I left it with "this works" on the version in the merged code.
https://stackoverflow.com/questions/35072192/how-to-make-make-append-additional-standard-flags

PR #308, opened 2021-05-04
Write t-route states to WRF-Hydro restart files

awlostowski-noaa:
New capability to write t-route simulated flow and depth data to WRF-Hydro restart files. The routine opens each WRF-Hydro restart file 

1. Construct a datetime array for the t-route simulation. This is needed to query t-route results against timestamps specified in WRF-Hydro Restart Files. 
2. Re-index the `flowveldepth` dataframe to match segment ID ordering in the Restart FIle
3. Open each WRF-Hydro restart file in a user-specified restart file directory with user specified file pattern filter
    1. Extract restart timestamp by converting Restart_Time attribute to datetime64 datatype
    2. Find the index of t-route datetime array that matches timestamp in Restart File. 
    3. IF a matching timestamp is found, THEN  create Data Arrays of t-route flow and depth data for the matching timestamp, add new variables to Restart File dataset, write edited data set to disk

### DIFF
--- a/src/python_framework_v02/troute/nhd_io.py
+++ b/src/python_framework_v02/troute/nhd_io.py
@@ -525,7 +525,98 @@ def get_channel_restart_from_wrf_hydro(
 
     return q_initial_states
 
+def write_channel_restart_to_wrf_hydro(
+    data,
+    restart_files,
+    channel_initial_states_file,
+    dt_troute,
+    nts_troute,
+    crosswalk_file,
+    channel_ID_column,
+    restart_file_dimension_var = 'links',
+    troute_us_flow_var_name = 'qlink1_troute',
+    troute_ds_flow_var_name = 'qlink2_troute',
+    troute_depth_var_name = 'hlink_troute',
+):
+    """
+    Write t-route flow and depth data to WRF-Hydro restart files. New WRF-Hydro restart
+    files are created that contain all of the data in the original files, plus t-route
+    flow and depth data. 
+    
+    Agruments
+    ---------
+        data (Data Frame): t-route simulated flow, velocity and depth data
+        restart_files (list): globbed list of WRF-Hydro restart files
+        channel_initial_states_file (str): WRF-HYDRO standard restart file used to initiate t-route simulation
+        dt_troute (int): timestep of t-route simulation (seconds)
+        nts_troute (int): number of t-route simulation timesteps
+        crosswalk_file (str): File containing reservoir IDs IN THE ORDER of the Restart File
+        channel_ID_column (str): field in the crosswalk file to assign as the index of the restart values
+        restart_file_dimension_var (str): name of flow and depth data dimension in the Restart File
+        troute_us_flow_var_name (str):
+        troute_ds_flow_var_name (str):
+        troute_depth_var_name (str):
+        
+    Returns
+    -------
+    """
+    # create t-route simulation timestamp array
+    with xr.open_dataset(channel_initial_states_file) as ds:
+        t0 = ds.Restart_Time.replace('_', ' ')
+    t0 = np.array(t0,dtype = np.datetime64)
+    troute_dt = np.timedelta64(dt_troute, 's')
+    troute_timestamps = t0 + np.arange(nts_troute) * troute_dt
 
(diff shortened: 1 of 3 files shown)


src/python_routing_v02/compute_nhd_routing_SingleSeg_v02.py
```diff
@@ -1286,7 +1286,37 @@ def main():
...
+                    output_parameters["wrf_hydro_channel_restart_pattern_filter"] + "[!.TROUTE]"
+                )
+            )  
+            
+            try:
+                if not pathlib.Path(restart_dir).exists() or len(restart_files) == 0:
+                    raise AssertionError
+            except AssertionError:
+                print('WRF Hydro restart files not found - Aborting restart write sequence')
+            else:
```
awlostowski-noaa, line 1307: This is a lame attempt at exception handling. I didn't want the whole workflow to fail if the restart files are not found. But maybe we want the whole process to fail?

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -469,7 +469,98 @@ def get_channel_restart_from_wrf_hydro(
...
+    for f in restart_files:
+        with xr.open_dataset(f) as ds:
+
+            # get timestamp from restart file
+            t = np.array(
+                ds.Restart_Time.replace('_', ' '), 
+                dtype = np.datetime64)
+
+            # find index troute_timestamp value that matches restart file timestamp
+            a = np.where(troute_timestamps == t)[0].tolist()
```
awlostowski-noaa, line 533: Unfortunately, I don't see a way around opening each restart file one at a time. This could be problematic if there hundreds of restart files, only a handful of which fall within the time domain of a t-route simulation. To avoid this, we might comprehend the restart time from the file name, but I am concerned that the file naming convention may not be as consistent as we'd like it to be.

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -469,7 +469,98 @@ def get_channel_restart_from_wrf_hydro(
...
+        
+    Returns
+    -------
+    """
+    # create t-route simulation timestamp array
+    with xr.open_dataset(channel_initial_states_file) as ds:
+        t0 = ds.Restart_Time.replace('_', ' ')
+    t0 = np.array(t0,dtype = np.datetime64)
+    troute_dt = np.timedelta64(dt_troute, 's')
+    troute_timestamps = t0 + np.arange(nts_troute) * troute_dt
```
awlostowski-noaa, line 512: Un until now, we have been working in the timestep space. That is, each column of the flowveldepth dataframe is indexed by a timestep, not a timestamp. In order to put the correct data into restart files, we need to work in the timestamp space.

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -469,7 +469,98 @@ def get_channel_restart_from_wrf_hydro(
...
+
+            # find index troute_timestamp value that matches restart file timestamp
+            a = np.where(troute_timestamps == t)[0].tolist()
+
+            # if the restart timestamp exists in the t-route simulatuion
+            if len(a) > 0:
+
+                # pull flow data from flowveldepth array, package into DataArray
+                # !! TO DO - is there a more percise way to slice flowveldepth array?
+                qtrt = flowveldepth_reindex.iloc[:,::3].iloc[:,a].to_numpy().astype("float32")
```
awlostowski-noaa, line 540: Would love a more robust way to index, here. I struggle with multi-indexed dataframes

jameshalgren:
Rebased and re-pushed (--force) after verifying with the author that updates were not broken by the rebase. Here is the short list of remaining technical items to complete before merging: 
- [ ] Possibly parameterize TROUTE output filename extension. 
- [ ] Make it consistent (or not, if intentional) with CHRTOUT extensions
- [ ] Slightly change the logic to trigger writing. 
- [x] Look forward isolated loop blocks... (Will wrap that into the looping PR)
- [ ] Ask NCAR --- why are there flow values under the reservoirs?

jameshalgren (approved):
Will need a performance review on a larger test dataset as a new piece of work.

PR #309, opened 2021-05-05
00 Routing package restructure

hellkite500:
Some refactoring of package structure and decoupling main entrypoint from routing/network backends.

## Additions

- nwm_routing package

## Changes

- next routing under troute namespace

## Notes

- See src/python_routing_v02/troute/readme.md

## Checklist


src/python_routing_v02/troute/README.md
```diff
@@ -0,0 +1,11 @@
+## Notes:
+Cython doesn't understand namespace packages, or if it does, I don't know how to make it use them.  To work around that, in the namespace directory, add `__init__.pxd`.  Python will ignore this and can treat the directory as a namespace package, and Cython will see it as a "normal" package and allow relative imports to work correctly.
+
+
+Framework and Routing are too dependent to really use two differnt pacakges, even under the same namespace.
+
+It would make sense to refactor the framework so that routing is a subpackage.  Then perhaps implement a subpackage which allows plugging in different network types and routing types?  Some mix of namespace package (for the framework) with plugin like utility.
```
jameshalgren, line 7: @hellkite500 
I might be missing the point but I think this chunk of code (which is not yet merged, but will be soon, hopefully --- we need to decide how best to juggle it with the present PR) describes the front end unit of network preprocessing: 
https://github.com/jameshalgren/t-route/blob/self_loop_merge2/src/python_routing_v02/compute_nhd_routing_SingleSeg_v02.py#L1459-L1474

Basically, that creates a topological ordering that could be saved as a static object and be route-able, if the required dynamic inputs (lateral inflows, temporal and spatial boundary conditions, data assimilation targets) are supplied. Perhaps that can guide the continuing effort to streamline this as a generalized package.

Continuing the review...

src/nwm_routing/setup.cfg
```diff
@@ -0,0 +1,20 @@
...
+
+[options]
+package_dir=
+  =src
+packages=find:
+python_requires=>=3.6
+install_requires=
+  numpy
+  Cython
+  pandas
```
groutr, line 14: IIRC, James ran into some issues using an older version of pandas. I believe the constraint here is pandas >= 1.2

hellkite500: Hmm....using python 3.6.8, pandas 1.1.5 is the latest available version.  We should really check into this in more detail if there is a bug or something that doesn't work with this version.  I haven't seen it yet, but doesn't mean I have hit it...For now, I'm going to revert the requirement to 1.1.5, we can adjust if needed, but that will mean a python 3.7.1 minimum if we want to use pandas 1.2 as the minimum.

jameshalgren:
@hellkite500 [This ](https://github.com/hellkite500/t-route/pull/2)fixes the general function failure using v03 execution. We still need to 
- [ ] add something so that v03 and v02 are not hardwired.
- [ ] black the new source files
- [ ] anything else

src/nwm_routing/setup.cfg
jameshalgren: So far, using the latest version of pandas on 3.6.8 (i.e. pandas 1.1.5) does not trigger the anomalous behavior so that should be a sufficient version requirement for now.

jameshalgren (approved):
Creating issue regarding courant failure. All else is solid.

hellkite500:
3f8c5680504ed3b74e43d770dd5c3423bd5a9ff8 put  compute_nhd_routing_SingleSeg_v02.py back in the index.  Will need to get it `rm`ed elsewhere, or could possible revert this PR, ammend it, and re-commit. @jameshalgren

PR #310, opened 2021-05-06
14 Add Nudging decay (including reading LastObs files)

jhrehanoaa:
creates the last obs dataframe. 
sets a true/false condition in the yaml to use either fresh start last obs file to create da decay predictions or run from our models final discharge values.
if no discharge values from routelink are matching it stop the run until this condition is fixed.
still need to generalize some of the code further.


src/python_framework_v02/troute/nhd_io.py
```diff
@@ -527,3 +528,69 @@ def get_reservoir_restart_from_wrf_hydro(
...
+            )
+            if time == 0:
+                prediction_df[str(time)] = model_discharge_last_ts["last_nudge"]
+                weight_diff = prediction_df[str(time)] - prediction_df[str(time)]
+            else:
+                if (weight > 0.1) == True:
+                    prediction_df[str(time)] = (
+                        delta["last_nudge"] + model_discharge_last_ts["model_discharge"]
+                    )
+                elif (weight < -0.1) == True:
```
groutr, line 591: This should be
```python
elif weight < -0.1:
```

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -527,3 +528,69 @@ def get_reservoir_restart_from_wrf_hydro(
...
+                print("THE NUDGING FILE IDS DO NOT MATCH THE FLOWVELDEPTH IDS")
+                sys.exit()
+        # Predictions created with continuously decreasing deltas until near 0 difference
+        a = 120
+        prediction_df = pd.DataFrame(index=model_discharge_last_ts.index)
+
+        for time in range(0, 720, 5):
+            weight = np.exp(time / -120)
+            delta = pd.DataFrame(
+                model_discharge_last_ts["last_nudge"] / np.exp(time / -120)
```
groutr, line 581: The value we computed as `weight` should be used here as the denominator.

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -527,3 +528,69 @@ def get_reservoir_restart_from_wrf_hydro(
...
+            model_discharge_last_ts["model_discharge"] != -9999.0
+        ]
+        model_discharge_last_ts = model_discharge_last_ts.reset_index().set_index(
+            "stationId"
+        )
+        model_discharge_last_ts = model_discharge_last_ts.drop(
+            ["stationIdInd", "timeInd"], axis=1
+        )
+        # If predict from last_obs file use last obs file results
+        if wrf_last_obs_flag == True:
```
groutr, line 559: `if wrf_last_obs_flag:`

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -527,3 +528,69 @@ def get_reservoir_restart_from_wrf_hydro(
...
+                )
+            else:
+                print("THE NUDGING FILE IDS DO NOT MATCH THE FLOWVELDEPTH IDS")
+                sys.exit()
+        # Predictions created with continuously decreasing deltas until near 0 difference
+        a = 120
+        prediction_df = pd.DataFrame(index=model_discharge_last_ts.index)
+
+        for time in range(0, 720, 5):
+            weight = math.exp(time / -120)
```
jameshalgren, line 579: Use the var...
```weight = math.exp(time / -a)```

File: src/python_routing_v02/compute_nhd_routing_SingleSeg_v02.py [hunk 5]
```diff
@@ -1197,6 +1197,19 @@ def main():
     else:
         usgs_df = pd.DataFrame()
 
+    last_obs_file = data_assimilation_parameters.get("wrf_hydro_last_obs_file", None)
+    last_obs_flag = data_assimilation_parameters.get("wrf_last_obs_flag", None)
+    if last_obs_file:
+        # fvd_df = flowveldepth.iloc[ :, -1:]
```
jameshalgren, line 1203: We will need to prepare the lastobs_df prior to calculation so that the effect can propagate through the network. So we cannot use the not-yet-created flowveldepth results.

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -700,3 +702,69 @@ def build_coastal_ncdf_dataframe(coastal_ncdf):
...
+        #     print("NO WRF_FLAG")
+        #     fvd_df = fvd_df
+        #     if len(model_discharge_last_ts.index) == len(fvd_df.index):
+        #         model_discharge_last_ts["last_nudge"] = (
+        #             model_discharge_last_ts["discharge"] - fvd_df[fvd_df.columns[0]]
+        #         )
+        #     else:
+        #         print("THE NUDGING FILE IDS DO NOT MATCH THE FLOWVELDEPTH IDS")
+        #         sys.exit()
+        # Predictions created with continuously decreasing deltas until near 0 difference
```
jameshalgren, line 750: See comment below about flowveldepth...

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -527,3 +528,69 @@ def get_reservoir_restart_from_wrf_hydro(
...
+                )
+            else:
+                print("THE NUDGING FILE IDS DO NOT MATCH THE FLOWVELDEPTH IDS")
+                sys.exit()
+        # Predictions created with continuously decreasing deltas until near 0 difference
+        a = 120
+        prediction_df = pd.DataFrame(index=model_discharge_last_ts.index)
+
+        for time in range(0, 720, 5):
+            weight = np.exp(time / -120)
```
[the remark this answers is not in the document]
jameshalgren: the weight is eventually applied to an array and it might be efficient to stack the computation into the array operation and perhaps take advantage of numpy optimizations. Just thinking...

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -700,3 +702,69 @@ def build_coastal_ncdf_dataframe(coastal_ncdf):
...
+        #         print("THE NUDGING FILE IDS DO NOT MATCH THE FLOWVELDEPTH IDS")
+        #         sys.exit()
+        # Predictions created with continuously decreasing deltas until near 0 difference
+        a = 120
+        prediction_df = pd.DataFrame(index=model_discharge_last_ts.index)
+
+        for time in range(0, 720, 5):
+            weight = math.exp(time / -a)
+            delta = pd.DataFrame(model_discharge_last_ts["last_nudge"] / weight)
+            if time == 0:
```
groutr, line 757: Perhaps the loop can be simplified by lifting the one special case outside the loop.

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -700,3 +702,69 @@ def build_coastal_ncdf_dataframe(coastal_ncdf):
...
+        #     else:
+        #         print("THE NUDGING FILE IDS DO NOT MATCH THE FLOWVELDEPTH IDS")
+        #         sys.exit()
+        # Predictions created with continuously decreasing deltas until near 0 difference
+        a = 120
+        prediction_df = pd.DataFrame(index=model_discharge_last_ts.index)
+
+        for time in range(0, 720, 5):
+            weight = math.exp(time / -a)
+            delta = pd.DataFrame(model_discharge_last_ts["last_nudge"] / weight)
```
groutr, line 756: Do we require a dataframe here? The call to pd.DataFrame I think is unnecessary.

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -700,3 +702,69 @@ def build_coastal_ncdf_dataframe(coastal_ncdf):
...
+            if time == 0:
+                prediction_df[str(time)] = model_discharge_last_ts["last_nudge"]
+                weight_diff = prediction_df[str(time)] - prediction_df[str(time)]
+            else:
+                if weight > 0.1:
```
groutr, line 761: `if abs(weight) > 0.1:`
???

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -700,3 +702,69 @@ def build_coastal_ncdf_dataframe(coastal_ncdf):
...
+        #         sys.exit()
+        # Predictions created with continuously decreasing deltas until near 0 difference
+        a = 120
+        prediction_df = pd.DataFrame(index=model_discharge_last_ts.index)
+
+        for time in range(0, 720, 5):
+            weight = math.exp(time / -a)
+            delta = pd.DataFrame(model_discharge_last_ts["last_nudge"] / weight)
+            if time == 0:
+                prediction_df[str(time)] = model_discharge_last_ts["last_nudge"]
```
groutr, line 758: maybe store `str(time)` to a variable some where.

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -700,3 +702,69 @@ def build_coastal_ncdf_dataframe(coastal_ncdf):
...
+            else:
+                if weight > 0.1:
+                    prediction_df[str(time)] = (
+                        delta["last_nudge"] + model_discharge_last_ts["model_discharge"]
+                    )
+                elif weight < -0.1:
+                    prediction_df[str(time)] = (
+                        delta["last_nudge"] + model_discharge_last_ts["model_discharge"]
+                    )
+        prediction_df["0"] = model_discharge_last_ts["model_discharge"]
```
groutr, line 769: Does this make checking time==0 redundant in the loop?

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -421,32 +510,42 @@ def get_usgs_from_time_slices_folder(
...
+    dates = []
+    for j in daterange(date_time_obj_start, date_time_obj_end):
+        dates.append(j.strftime("%Y-%m-%d_%H:%M:00"))
+
+    for i, j in enumerate(dates):
+        if usgs_df.iloc[:, i].name == dates[i]:
+            pass
+        else:
+            if usgs_df.iloc[:, i].name not in dates:
+                usgs_df.drop(usgs_df.columns[i], axis=1, inplace=True)
```
jameshalgren, line 537: ping @groutr. We discovered that some of the time stamps in the USGS time slices files are unexpected -- there are some that fall on 5 minute marks, which is sort-of ok, and there are some that fall on sub-five-minute intervals. As we explored the datasets, it appears that the data in the for example 15:31 timestamp is identical to the 15:30 timestamp. SO... we are deleting what appear to be duplicated entries and interpolating over them.

jameshalgren:
- [ ] Fix Diffusive function signature to include last_obs
(test with `python compute_nhd_routing_SingleSeg_v02.py -f ../../test/input/yaml/Florence_Benchmark_diff.yaml`)

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -338,6 +340,92 @@ def preprocess_time_station_index(xd):
...
+        #         if weight > 0.1:
+        #             prediction_df[str(time)] = (
+        #                 delta["last_nudge"] + model_discharge_last_ts["model_discharge"]
+        #             )
+        #         elif weight < -0.1:
+        #             prediction_df[str(time)] = (
+        #                 delta["last_nudge"] + model_discharge_last_ts["model_discharge"]
+        #             )
+        # prediction_df["0"] = model_discharge_last_ts["model_discharge"]
+        return final_df
```
jameshalgren, line 427: @jhrehanoaa 
- [ ] Need to return last_obs_df *and* last_obs_time_df.

jameshalgren (approved):

PR #322, opened 2021-05-28
DA Decay NCAR 

jhrehanoaa:
This PR incorporates everything we have been trying to accomplish with da. 

- Performs da substitutions when necessary
- Loads values from last obs file
- Starts ncar decay method once running out of timeslice files

There is now the capability to locate at what time index the last observation file will occur. This index value is passed to mc_reach as last_obs_start. mc_reach will perform the normal da substitutions until running out of timeslice files. Naturally, this is when the last_obs file is read and substituted. From this point on, da using ncars method of delta calculation is performed. Currently, the last obs file is the first-time step, so no substitution is occurring. However, this can be changed by loading da timeslice files up to a point where there is a last obs file. We have the yaml set to stop before the number of qlat files in order to demonstrate we could decay appropriately when the last obs file read capability was turned off.


src/python_framework_v02/troute/nhd_network.py
```diff
@@ -88,6 +88,31 @@ def reverse_network(N):
...
+        [ ... ] ] reach n
+    tw 2:
+        etc.
+    '''
+    for tw, rs in reaches_bytw.items():
+        for r in rs:
+            if node in r:
+                return(tw)
+
+    return None  # Node not in reach set.
```
groutr, line 113: It's likely better here to raise an exception instead of returning None.

jhrehanoaa:
Replaced by another PR

PR #333, opened 2021-06-22
Diffusive code update into framework

jameshalgren:


src/fortran_routing/diffusive_v02frwk/maindiff_v2.f90
```diff
@@ -0,0 +1,1827 @@
...
+            ncomp= frnw_g(j,1)
+            do i = 1,ncomp-1
+                dx(i,j) = dx_ar_g(i,j)
+            end do
+            minDx = min(minDx,minval(dx(1:ncomp-1,j)))
+        end do
+
+        z=z_ar_g
+        ini_y=0.05  !* [meter]
+        ini_q=0.1   !*[m^3/sec]
```
jameshalgren, line 225: These will need to be inputs generated from a restart file.

jameshalgren, line 225: These will need to be inputs generated from a restart file.

jameshalgren (commented):
need to add compatibility with restart files

PR #334, opened 2021-06-22
DA structured

jhrehanoaa:
This incorporates the code used to get the new computation method of structured-obj and structured with DA decay. Currently, it operates correctly in structured-obj and structured, but no longer in caching due to a new buffer issue created during the overhaul of the new packaging scheme.

This pull requests suggest three contributions to the codebase:

Data assimilation (no decay) compatibility with structured compute method
The basic workflow of replacing flow simulations with gage observations is borrowed from the mc_reach.compute_network method and implemented in mc_reach.compute_network_structured. This is an important contribution because data assimilation is otherwise not possible in mc_reach.compute_network_structured. Moreover, mc_reach.compute_network_structured is the only compute scheme capable of handling reservoirs. This update gives us the capability to include both reservoirs and DA into a routing simulation.


jhrehanoaa:
Closing in order to open a new PR with reworked DA capabilities.

PR #335, opened 2021-06-25
DA Rework Last_Obs

jhrehanoaa:
Helps provide us a test case for the reworking of DA. Gaps are still interpolated. 

Changed some of the code to rework how last_obs are read and decay is performed. Will start decaying immediately if last_obs is provided in structured and structured-obj. If one is not provided then substitution is performed naturally. Should decay begin after the end of substitutions? 

This incorporates the code used to get the new computation method of structured-obj and structured with DA decay. Currently, it operates correctly in structured-obj and structured, but no longer in caching due to a new buffer issue created during the overhaul of the new packaging scheme.

This pull requests suggest three contributions to the codebase:

Data assimilation (no decay) compatibility with structured compute method
The basic workflow of replacing flow simulations with gage observations is borrowed from the mc_reach.compute_network method and implemented in mc_reach.compute_network_structured. This is an important contribution because data assimilation is otherwise not possible in mc_reach.compute_network_structured. Moreover, mc_reach.compute_network_structured is the only compute scheme capable of handling reservoirs. This update gives us the capability to include both reservoirs and DA into a routing simulation.


jhrehanoaa:
replaced by another PR

Issue #337, opened 2021-06-29
Update documentation


jameshalgren:
The documentation should be updated to reflect the following installation/execution steps: 
``` sh

# make a clean python virtual environment
cd ~
cd ..
mkdir venvs && cd venvs
python3 -m venv py36_clean
source py36_clean/bin/activate

cd ~
mkdir git && cd git
git clone https://github.com/[your github username]/t-route t-route-clean
cd t-route-clean/
git remote add upstream https://github.com/NOAA-OWP/t-route
git fetch upstream
git checkout upstream/master
git remote add upstream https://github.com/NOAA-OWP/t-route

cd src/nwm_routing/

# If python>=3.7
    pip install hydrotools.nwis_client

# ElseIf python==3.6
    pip install --upgrade pip  # If you are using python 3.6, you probably need to upgrade pip
    pip install wheel
    pip install hydrotools.nwis-client==2.0.0a0
    pip install hydrotools.-restclient==2.0.0a0

pip install -e .
pip install dask

cd src/python_routing_v02
./compiler.sh

python -m nwm_routing -V2 -f ../../test/input/yaml/Florence_Benchmark_diff.yaml

```

jameshalgren:
The IVDataService call in diffusive_utils.py was incompatible with the newer version of the nwis-client needed for compatibility with Python 3.7. #338 addresses this incompatibility, but the notes above are still relevant.

PR #338, opened 2021-06-29
Update IVDataService call

jameshalgren:
This micro PR converts to a new usage: we create an instance of IVDataService and call the get method of that instance. 

The get method previously was a classmethod, but in the new version it is not. This change was to support better caching and other optimizations. The good news is, the new syntax is backward compatible to the previous version of the nwis client so no change is needed to any existing environments with this update, but now we can support any newer environments. Please see #337 for additional information.

**Note:** The dependency on NWIS client access within `diffusive_utils.py` is probably non-optimal (placement within `nhd_io.py`, for instance, seems more appropriate). This PR is focusing on an issue that prevents migrating to newer environments, needed for asynchronous data loading capability. Further changes to sort out the placement of the data call can be handled in a later PR.

### DIFF
--- a/src/python_routing_v02/troute/routing/diffusive_utils.py
+++ b/src/python_routing_v02/troute/routing/diffusive_utils.py
@@ -326,7 +326,7 @@ def fp_ubcd_map(frnw_g, pynw, nts_ub_g, nrch_g, geo_index, qlat_data, qlat_g):
     return ubcd_g
 
 
-def fp_dbcd_map(usgsID2tw, usgssDT, usgseDT, usgspCd):
+def fp_dbcd_map(usgsID2tw=None, usgssDT=None, usgseDT=None, usgspCd=None):
     """
     Downststream boundary condition mapping between Python and Fortran using USGS stage observations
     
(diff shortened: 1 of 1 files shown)


awlostowski-noaa:
As of now, I think the creation of `bdcd_g`, the downstream boundary condition array, is contingent on getting USGS gage observations. Can we increase the scope of this PR to drop that requirement? This is a barrier to diffusive integration.

awlostowski-noaa:
Answering my own questions from above...

Here are the lines of Fortran code in `diffusive.f90` that handle the downstream boundary condition data in `dbcd_g`:
```fortran
                    if (dbcd_g(1)>0.0) then
                    !* measured or equivalent data available at TW node
                        do n=1,nts_db_g
                            varr_db(n)= dbcd_g(n)
                        enddo
                        tf0= tc+dtini_g/60.0
                        elv_g(ncomp,j)= intp_y(nts_db_g, tarr_db, varr_db, tf0)
                    else
                    !* when no measured data avail, use normal depth instead
                        dsc=abs(q_g(ncomp,j))
                        z_g= z_ar_g(ncomp,j)
                        bo_g= bo_ar_g(ncomp,j)
                        traps_g= traps_ar_g(ncomp,j)
                        So_g= so_ar_g(ncomp,j)
                        tw_g= tw_ar_g(ncomp,j)
                        twcc_g= twcc_ar_g(ncomp,j)
                        mann_g= mann_ar_g(ncomp,j)  !1.0/sk(i,j)
                        manncc_g= manncc_ar_g(ncomp,j) ! 1.0/skCC1(i,j)
                        hbf= bfuf_g(ncomp,j,1)      !* bankfull water depth
                        ufQbf= bfuf_g(ncomp,j,2)    !* bankfull uniform flow


                        call bsec_nmdep(dsc, hbf, ufQbf, hnm)
```
So, seems like passing `dbcd_g` array of zeros will enable a normal depth downstream boundary calculation.

awlostowski-noaa (commented):
```python
def fp_dbcd_map(usgsID2tw, usgssDT, usgseDT, usgspCd):
```
Can we make these inputs optional? e.g. `usgsID2tw = None`. Then, we can place an if statement that will only use the `nwis_client` if it parameters are specified.

kumdonoaa:
Let's make the last line: dbcd_g=np.zeros(nts_db_g) -> dbcd_g= -np.ones(nts_db_g)   
So, all -1.

awlostowski-noaa:
@kumdonoaa good idea. I'm making these changes in a pull request on James's branch, here.

PR #339, opened 2021-06-30
Output cleanup

jameshalgren:
Required output block was missing. Made a few additional minor updates.

### DIFF
--- a/src/nwm_routing/src/nwm_routing/__main__.py
+++ b/src/nwm_routing/src/nwm_routing/__main__.py
@@ -567,13 +567,12 @@ def main_v02(argv):
     if verbose:
         print(f"Handling output ...")
 
-    if output_parameters:
-        csv_output = output_parameters.get("csv_output", None)
-        if csv_output:
-            csv_output_folder = output_parameters["csv_output"].get(
-                "csv_output_folder", None
-            )
-            csv_output_segments = csv_output.get("csv_output_segments", None)
+    csv_output = output_parameters.get("csv_output", None)
+    if csv_output:
+        csv_output_folder = output_parameters["csv_output"].get(
+            "csv_output_folder", None
+        )
+        csv_output_segments = csv_output.get("csv_output_segments", None)
 
     if (debuglevel <= -1) or csv_output:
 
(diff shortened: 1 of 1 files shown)


jameshalgren:
@rcabell

src/nwm_routing/src/nwm_routing/__main__.py
```diff
@@ -634,6 +627,76 @@ def main_v02(argv):
...
+                    "wrf_hydro_channel_ID_crosswalk_file_field_name"
+                ),
+                wrf_hydro_channel_restart_new_extension,
+            )
+        else:
+            # print error and/or raise exception
+            print(
+                "WRF Hydro restart files not found - Aborting restart write sequence"
+            )
+            raise AssertionError
```
rcabell, line 676: Is there a reason not to use the `raise Assertion(message: str)` construct here instead of printing the message separately?

awlostowski-noaa: No reason - I just didn't know how to correctly use `raise AssertionError` when I first designed this. Thanks for noticing this!

awlostowski-noaa (approved):

src/nwm_routing/src/nwm_routing/__main__.py
jameshalgren: @rcabell -- like this?
``` python
else:
    # print error and raise exception
    str = "WRF Hydro restart files not found - Aborting restart write sequence"
    raise AssertionError(str)
```

rcabell: Yes, that's the idea. You can pass the string in directly, too. By doing this you get two advantages (beyond simpler code): the error will print to stderr vs stdout, and if a outer `except` block handles the error, it can inspect the message and print it / log it / do whatever with it.

jameshalgren:
@awlostowski-noaa -- I merged this with one slight additional change on your prior approval.
@rcabell -- thank you for the suggestion; this is now in the master. It doesn't fix the library issues we identified ---  see #342 for follow-up on that issue.

PR #344, opened 2021-07-02
Reservoir splitting subnetwork fix

### LINKED ISSUE #343: DFS network splitting function hits a snag when used with reservoirs
The `nhd_network.dfs_decomposition_depth_tuple` function, when applied in the up-and-coming diffusive-with-reservoirs application uses `split_at_junctions` as the `path_func` but should use `split_at_waterbodies_and_junctions` as the `path_func`. With `split_at_junctions`, the function returns an order of negative 1 for a network consisting of a single segment with an upstream that is a reservoir. 

e.g.
```
> from functools import partial
> rconn = {"down":["up (is a reservoir)"]}
> path_func = partial(nhd_network.split_at_junction, rconn)
> nhd_network.dfs_decomposition_depth_tuple(rconn, path_func)
[(-1, ['down'])]
```
While using what should be the appropriate splitting function yields an error:
```
> path_func = partial(nhd_network.split_at_waterbodies_and_junctions, ["up (is a reservoir)"], rconn)
> nhd_network.dfs_decomposition_depth_tuple(rconn, path_func)
*** TypeError: 'NoneType' object is not subscriptable
``` 

What is needed is probably a small mod to that function or pair of functions. 

In the regular dfs function `dfs_decomposition`, the splitting function (applied as a partial, `path_func`) is used only on the reverse traversal, i.e., after the depth-first search has reached a leaf of the tree. As it pops items out of the stack, the path variable keeps track of the values coming down the line.
As this happens, the splitting function examines both the node and the path. If the path is a reservoir and the node is not OR if the node is a reservoir and the path is not (i.e., if we are entering or leaving a reservoir), then the function is triggered and the path splits.

The problem is that in the slightly more complex `dfs_decomposition_depth_tuple`, I am trying to use that same splitting function in reverse on the way up the tree to increment a count of the reach depth... [this code](https://github.com/NOAA-OWP/t-route/blob/master/src/python_framework_v02/troute/nhd_network.py#L246-L248).

There is not a "path" defined at that point. So I was passing None to the function -- which works with the junction only splitting function (which doesn't even reference the path argument...) -- but the reservoir splitting function needs the path to detect the change.

_Possible Solutions_
* The `(order, reach)` tuples returned from the decomposition could be made by coalescing the reaches (e.g., in the `reaches_bytw` object and counting the depth. 
* The if statement could be rebuilt to accumulate the logic from the collection of potential children:
```
is_new_split = False
for _child in N[node]:  # list of upstreams
    if not path_func([node], _child):
        is_new_split = True  # if any of the children are a split, then they all should be.
        break
if is_new_split:
    if node in junctions:
        is_new_split = False
```

jameshalgren:
May close #343 -- this PR implements one of the suggested fixes to that issue. 
@jdmattern-noaa

### DIFF
--- a/src/python_framework_v02/troute/nhd_network.py
+++ b/src/python_framework_v02/troute/nhd_network.py
@@ -259,11 +259,11 @@ def dfs_decomposition_depth_tuple(RN, path_func, source_nodes=None):
             raise AssertionError(
                 "the source nodes *must* be members of the coalesced set..."
             )
-    depth_tuples = dfs_count_depth_coalesced(RN_coalesced, source_nodes)
+    depth_tuples = dfs_count_depth(RN_coalesced, source_nodes)
     return zip(pluck(0,depth_tuples), reach_list)
 
 
-def dfs_count_depth_coalesced(RN, source_nodes=None):
+def dfs_count_depth(RN, source_nodes=None):
     """
     Returns tree depth, via depth-first-search of every element
     in a network.
--- a/src/python_framework_v02/troute/nhd_network.py
+++ b/src/python_framework_v02/troute/nhd_network.py
@@ -95,7 +95,7 @@ def find_tw_for_node(reaches_bytw, node):
     """
     reaches_bytw is a dictionary of lists of the form
 
-    tw 1: 
+    tw 1:
       [ [ seg1, seg2, seg3, ... segn ], # reach 1
         [ sega, segb, segc, ... segz ], # reach 2
         .
--- a/src/python_framework_v02/troute/nhd_network.py
+++ b/src/python_framework_v02/troute/nhd_network.py
@@ -257,18 +257,18 @@ def dfs_decomposition_depth_tuple(RN, path_func, source_nodes=None):
             raise AssertionError(
                 "the source nodes *must* be members of the coalesced set..."
             )
-    depths = count_depth(RN_coalesced, source_nodes)
+    depths = dfs_count_depth(RN_coalesced, source_nodes)
     return zip(depth, reach_list)
 
 
-def count_depth(RN, source_nodes=None):
+def dfs_count_depth(RN, source_nodes=None):
 
     path_tuples = []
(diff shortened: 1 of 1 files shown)


jameshalgren:
Referencing the comment in #344, the output now is: 
```
> from functools import partial
> rconn = {"down":["up (is a reservoir)"]}
> path_func = partial(nhd_network.split_at_junction, rconn)
> nhd_network.dfs_decomposition_depth_tuple(rconn, path_func)
[(0, ['down'])]
```

jameshalgren:
This implements a dfs to accomplish the depth counting. 
A bfs approach might be 
``` python
level = 0
stack = deque([h])
while stack:
    level_size = len(stack)
    while level_size:
        node = stack.popleft()
        stack.extend(N[node])
        level_size -= 1
    level += 1
```
(there might be a bug or two in there, but the general idea is that we change the way we use the stack...we use it as a queue instead)

groutr:
> 
> 
> This implements a dfs to accomplish the depth counting.
> A bfs approach might be
> 
> ```python
> level = 0
> stack = deque([h])
> while stack:
>     level_size = len(stack)
>     while level_size:
>         node = stack.popleft()
>         stack.extend(N[node])
>         level_size -= 1
>     level += 1
> ```
> 
> (there might be a bug or two in there, but the general idea is that we change the way we use the stack...we use it as a queue instead)

The nice side-effect of this approach is that when the inner loop is entered, the first n elements of the stack are all on the same level. You could build a map of the level number and the set of nodes in that level.
```
levels = {}
level_counter = 0
Q = deque([h])
while Q:
    level_size = len(Q)
    tmp = set()
    while level_size > 0:
        node = Q.popleft()
        Q.extend(N[node])
        tmp.add(node)
        level_size -= 1
    levels[level_counter] = tmp
    level_counter += 1
```

jameshalgren:
superseded by #347

PR #357, opened 2021-07-28
Interior terminal nodes

jameshalgren:
Somewhere, way back, we had some sort of catch for a case where a set of masked segments that were part of a larger network would generate a valid connections set by setting the off-network terminal nodes to negative values. These 'to' or 'downstream' segment ids exist in the larger network, but when masked, are not part of the computation domain. Somewhere along the way, we lost that catch, but since all of our test cases had been boiled down to situations where the terminal codes are all 0, apparently. Or we have been just lucking-out. 

Either way, this introduces a more clear usage of the "terminal_code" and modifies the principal call in the code to this function to use the code and any non-standard internal codes in the call to the function `extract_connections`. 

~~The PR is frozen as a draft until I can add a unit test.~~

A Unit Test is added as part of the PR.

### DIFF
--- a/src/python_framework_v02/troute/nhd_network.py
+++ b/src/python_framework_v02/troute/nhd_network.py
@@ -42,7 +42,7 @@ def out_degrees(N):
     return in_degrees(reverse_network(N))
 
 
-def extract_connections(rows, target_col, terminal_code=0):
+def extract_connections(rows, target_col, terminal_codes={0}):
     """Extract connection network from dataframe.
 
     Arguments:
@@ -58,7 +58,7 @@ def extract_connections(rows, target_col, terminal_code=0):
         if src not in network:
             network[src] = []
 
-        if dst > 0:
+        if dst not in terminal_codes:
             network[src].append(dst)
     return network
 
--- a/src/python_framework_v02/troute/nhd_network.py
+++ b/src/python_framework_v02/troute/nhd_network.py
@@ -63,7 +63,7 @@ def extract_connections(rows, target_col, terminal_codes={0}):
     return network
 
 
-def extract_waterbodies(rows, target_col, waterbody_null=-9999):
+def extract_waterbody_connections(rows, target_col, waterbody_null=-9999):
     """Extract waterbody mapping from dataframe."""
     return (
         rows.loc[rows[target_col] != waterbody_null, target_col].astype("int").to_dict()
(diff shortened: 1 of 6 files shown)


src/python_framework_v02/troute/nhd_network.py
```diff
@@ -42,7 +42,7 @@ def out_degrees(N):
     return in_degrees(reverse_network(N))
 
 
-def extract_connections(rows, target_col, terminal_code=0):
+def extract_connections(rows, target_col, terminal_codes={0}):
```
groutr, line 45: Something like below would be a more defensive way and not require the user to worry to about what type of collection is passed. If the user doesn't pass an iterable, an error will be raised when the set is created.

```python
def extract_connections(rows, target_col, terminal_codes=None):
    if terminal_codes is not None and isinstance(terminal_codes, Iterable):
        terminal_codes = set(terminal_codes)
    else:
        terminal_codes = set()
```

groutr:
@jameshalgren I think you were just lucking out. The function hasn't changed very little since it was first written and was designed to catch the 0 terminal node only.

jameshalgren:
Changed the default input to None, to avoid the problem of having an immutable object as a default argument in a function (which is technically safe if the object is not modified, as in our case, but which is bad practice because of the potential for weird behavior if the object is modified.

src/python_framework_v02/troute/nhd_network_utilities_v02.py
```diff
@@ -445,7 +445,16 @@ def build_connections(supernetwork_parameters):
...
+    # There can be an externally determined terminal code -- that's this first value
+    terminal_codes = set()
+    terminal_codes.add(terminal_code)
+    # ... but there may also be off-domain nodes that are not explicitly identified
+    # but which are terminal (i.e., off-domain) as a result of a mask or some other
+    # an interior domain truncation that results in a 
+    # otherwise valid node value being pointed to, but which is masked out or 
+    # being intentionally separated into another domain.
+    terminal_codes = terminal_codes | set(param_df[~param_df["downstream"].isin(param_df.index)]["downstream"].values)
+    connections = nhd_network.extract_connections(param_df, "downstream", terminal_codes)
```
groutr, line 457: I try to pass keyword arguments explicitly.
`extract_connections(param_df, "downstream", terminal_codes=terminal_codes)`

groutr (approved):

Issue #368, opened 2021-08-04
Make the reach list tagging a function


jameshalgren:
These lines are from the `serial` execution pathway. 
https://github.com/NOAA-OWP/t-route/blob/70e32677509b3e767e1eeeb48be9bb1a76b15a41/src/python_routing_v02/troute/routing/compute.py#L997-L1013

Those lines are more compact than the other places where we do what is shown in the comments from the same block. 

So... let's create a simple helper function and reference the helper function in each location where we generate `reaches_list_with_type`

jameshalgren:
closed by #372

PR #372, opened 2021-08-06
replace reach list with type creation with a single common function

awlostowski-noaa:
Addressing #368

### DIFF
--- a/src/python_framework_v02/troute/nhd_network_utilities_v02.py
+++ b/src/python_framework_v02/troute/nhd_network_utilities_v02.py
@@ -679,3 +679,12 @@ def build_data_assimilation_folder(data_assimilation_parameters):
         )
 
     return usgs_df
+
+def build_reach_type_list(reach_list, wbodies_segs):
+    
+    reach_type_list = [
+                1 if (set(reaches) & wbodies_segs) else 0 for reaches in reach_list
+            ]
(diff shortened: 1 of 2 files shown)


src/python_framework_v02/troute/nhd_network_utilities_v02.py
```diff
@@ -679,3 +679,12 @@ def build_data_assimilation_folder(data_assimilation_parameters):
         )
 
     return usgs_df
+
+def build_reach_type_list(reach_list, wbodies_segs):
+    
+    reach_type_list = [
+                1 if (set(reaches) & wbodies_segs) else 0 for reaches in reach_list
+            ]
+    reaches_list_with_type = list(zip(reach_list, reach_type_list))
```
jameshalgren, line 688: Use `return list(zip(reach_list, reach_type_list))` -- no need to allocate the extra variable.

jameshalgren (changes requested):
One other possibility, but not a blocker for merging: we could move the function into compute.py and name it 
`_build_reach_type_list`

I don't see very many (any?) use cases where we would use this function outside of the compute function.

jameshalgren:
and @awlostowski-noaa thank you!

jameshalgren (approved):
passes all prior tests.

PR #373, opened 2021-08-09
cython indentation fix

jhrehanoaa:
Applied a simple file formatter to the cython source to improve the diffs. Similarly to black we can use this in the command line to fix all the cython files with indent errors. 

pip install reindent 
move to directory
reindent {filename}


jameshalgren (approved):
Thank  you --- just right.

Issue #376, opened 2021-08-09
Reuse of 'i' creates potential indexing collision


jameshalgren:
These two loops modify the main loop index `i` and should be changed to use `_i` 

https://github.com/NOAA-OWP/t-route/blob/341fed06fae03e39a9def8d9165ac47565264492/src/python_routing_v02/troute/routing/fast_reach/mc_reach.pyx#L1246

https://github.com/NOAA-OWP/t-route/blob/341fed06fae03e39a9def8d9165ac47565264492/src/python_routing_v02/troute/routing/fast_reach/mc_reach.pyx#L1268

as an example, the `_i` syntax was used here: 
https://github.com/NOAA-OWP/t-route/blob/341fed06fae03e39a9def8d9165ac47565264492/src/python_routing_v02/troute/routing/fast_reach/mc_reach.pyx#L1218

PR #377, opened 2021-08-10
fixed i indexing collisions in mc_reach

jhrehanoaa:
Fixes issue Reuse of 'i' creates potential indexing collision #376.


jameshalgren:
Rolled into #358

Issue #362, opened 2021-08-03
Courant = true breaks Florence_Benchmark.yaml test case


jameshalgren:
`return_courant: true` breaks execution for the Florence_Benchmark.yaml test case -- this is because the `compute_network_structured` compute kernel does not produce any valid values for courant output and later logic then fails on the empty output. The failure can be tested with this command line: 
```python -m nwm_routing -V2  -f ../../test/input/yaml/Florence_Benchmark.yaml```
Setting `return_courant: false` for the failed test case works. 

Other work (#380) has fixed the courant return for the `compute_network` kernel (indicated in the yaml with `compute_kernel: V02-caching`) but porting the fix to the other kernels requires handling of the reservoirs, which has not yet been prototyped. A simple possibility would be to populate a Nan or Zero array for all Courant returns from reservoirs. That would have implications for later evaluation but may be sufficient to proceed. 

_Originally posted by @jhrehanoaa in https://github.com/NOAA-OWP/t-route/issues/347#issuecomment-880906590_

jameshalgren:
@awlostowski-noaa 
We may need to fix this soon.

jameshalgren:
at line 593 in  __main__py
`pd.concat([pd.DataFrame(r[2], index=r[0], columns=pd.MultiIndex.from_product([range(8641), ["cn", "ck", "X"]]).to_flat_index()) for r in results], copy=False,)`
`pd.concat([pd.DataFrame(r[1], index=r[0], columns=pd.MultiIndex.from_product([range(8640), ["q", "v", "d"]]).to_flat_index()) for r in results], copy=False,)`

jameshalgren:
The issue was that we had added (then subtracted) a timestep to the main array but not to the courant array.

jameshalgren:
#362 fixes this for the V02_caching mode. This does not yet fix the broken test python -m nwm_routing -V2 -f ../../test/input/yaml/Florence_Benchmark.yaml because the yaml file refers to the V02-structured compute method, which does not yet handle the courant output properly.

JurgenZach-NOAA:
Obsolete test case, but issue addressed in PR380

PR #380, opened 2021-08-12
[WIP] Courant timestep bugfix

jameshalgren:
Bugfix to return capability for returning courant number for simulations using the V02-caching compute kernel. Per comments in issue #362 the problem was that an additional index position was added to accomodate initial conditions in the main array but not in the courant array. This PR now makes the behavior for the courant array match the flowveldepth array. We could probably avoid this somehow, but this seems to work for now. 

This does not yet fix the broken test `python -m nwm_routing -V2  -f ../../test/input/yaml/Florence_Benchmark.yaml` because the yaml file refers to the `V02-structured` compute method, which does not yet handle the courant output properly.

- [ ] Annotate issue to note need to add extra methods

### DIFF
--- a/src/nwm_routing/src/nwm_routing/__main__.py
+++ b/src/nwm_routing/src/nwm_routing/__main__.py
@@ -597,8 +597,8 @@ def main_v02(argv):
             ).to_flat_index()
             courant = pd.concat(
                 [
-                    pd.DataFrame(c, index=i, columns=courant_columns)
-                    for i, d, c in results
+                    pd.DataFrame(r[2], index=r[0], columns=courant_columns)
+                    for r in results
                 ],
                 copy=False,
             )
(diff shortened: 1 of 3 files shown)


jameshalgren:
ping @BrianAvant-NOAA

BrianAvant-NOAA:
@jameshalgren is this still in progress or should I be able to generate courant numbers? I checked this branch using `CustomInput.yaml` and `Florence_Benchmark.yaml` as inputs. If I set `return_courant` to false I can run both tests using the `V02-structured` or `V02-caching`  `compute_method`/`compute_kernel`. Any attempts setting `return_courant` to true still fail though.

BrianAvant-NOAA:
@jameshalgren just wanted to update on this ticket that we resolved the error I was having by re-compiling with the new branch. I am able to produce Courant numbers now for both test datasets.

jameshalgren:
> @jameshalgren just wanted to update on this ticket that we resolved the error I was having by re-compiling with the new branch. I am able to produce Courant numbers now for both test datasets.

Thank you for the update.

awlostowski-noaa (approved):

PR #385, opened 2021-08-18
Da update for caching structured obj

jameshalgren:
pr #358 left a few items incomplete on the `V02-caching` and `V02-structured-obj` computation modes. Those are not target operational modes so the PR was merged in spite of that. 

This PR fixes those items (using the loop-less da procedure; using `_i` for temporary indexes) and also updates the `V02-structured` DA algorithm with a minor but important update to the timestep indexing. 

The output parity is now slightly better: 
image

### DIFF
--- a/src/python_routing_v02/troute/routing/fast_reach/simple_da.pyx
+++ b/src/python_routing_v02/troute/routing/fast_reach/simple_da.pyx
@@ -1,6 +1,7 @@
 from libc.math cimport exp, isnan
 # from libc.stdio cimport printf
 
+# TODO: remove unused math functions from mc_reach.pyx
 
 cpdef float simple_da_with_decay_py(
     const float last_valid_obs,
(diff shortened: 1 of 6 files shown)


jameshalgren:
This PR implements three sections of nearly identical code in lines 455-473 and lines 1377-1430 in mc_reach.pyx. I'm thinking that we can remove some of the code duplication and hopefully streamline things (both generally desirable goals) for all the inputs if we create a function to handle those calculations. 

The challenge would be that there are slight differences in the handling of timesteps and in the declaration of the flowveldepth arrays. Can anyone think of a way to make a cython function to treat a cdef float[:,::1] and a cdef np.ndarray[float, ndim=3] as synonymous? We have those two constructions in our different compute functions representing the same variables (flowveldepth) and I would like to prepare a single function to initialize them both. In other words, I would like to create a function into which I can pass either the memoryview with 2 dimensions or the nd array with 3 dimensions and, given the congruency of their data storage, populate them equivalently.

One option would be to modify the caching-based version of our function to use the 3-dimensional declaration (so they would all be synchronous.) But that loses the advantages of the typed memoryview, which are more general and more efficiently used with cython. See the following: 

> "Typed memoryviews allow efficient access to memory buffers, such as those underlying NumPy arrays, without incurring any Python overhead. Memoryviews are similar to the current NumPy array buffer support (np.ndarray[np.float64_t, ndim=2]), but they have more features and cleaner syntax."
https://cython.readthedocs.io/en/latest/src/userguide/memoryviews.html

With that in mind, I'd prefer to leave the memoryview in place and find a way to convert the array declaration, at least for purposes of handling inside this function.

jameshalgren:
ping @groutr

src/python_routing_v02/troute/routing/fast_reach/simple_da.pyx
```diff
@@ -19,6 +20,63 @@ cpdef float simple_da_with_decay_py(
...
+        #     printf("a: %g\t", a)
+        #     printf("min: %g\t", da_decay_minutes)
+        #     printf("lo_ts: %d\t", lastobs_timestep[gage_i])
+        #     printf("ndg: %g\t", da_weighted_shift)
+        #     printf("lov: %g\t", lastobs_val)
+        #     printf("orig: %g\t", model_val)
+        #     printf("new: %g\t", replacement_val)
+        #     printf("\n")
+        return replacement_val
+        # return nudge, replacement_val
```
jameshalgren, line 77: @groutr The goal is to actually return these two values, not just the one. We could return the nudge only, then do the arithmetic (replacement_val = model_val + nudge) in the calling code.

jhrehanoaa (approved):

src/python_routing_v02/troute/routing/fast_reach/simple_da.pyx
jameshalgren: Resolved. See use of ctuple to declare function in simple_da.pyx:
``` py
cdef (float, float, float, float) simple_da(
...
```
... and the buffer to receive the results in mc_reach.pyx:
``` py
cdef (float, float, float, float) da_buf
...
    da_buf = simple_da(
        timestep,
...
```

jhrehanoaa (approved):

jhrehanoaa:
Works on Cheyenne with the fcast test case.

PR #393, opened 2021-08-24
Add MC kernel unit test

jameshalgren:
TL;DR Added a unit test and a more comprehensive test script to demonstrate parity between kernels. But... under certain conditions, parity is not demonstrated!

Execute
`src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/mc_sseg_stime_NOLOOP_demo.py` for a simple parity test printed to the screen. This shows that the SUBMUSKINGCUNGE function copied directly out of WRF-Hydro here produces the same kernel result for a pile of random inputs as our kernel. 
``` py
First test low-flow, showing expected result depending on precision of calculation.
real double precision expected q: 0.7570107902354513 vel: 0.12373606306742324 depth: 0.02334451646521419
real single precision expected q: 0.7570106983184814 vel: 0.12373604625463486 depth: 0.02334451675415039
real single precision computed via updated method q: 0.7570106983184814 vel: 0.12373604625463486 depth: 0.02334451675415039
real single precision computed via WRF-Hydro method q: 0.7570106983184814 vel: 0.12373604625463486 depth: 0.02334451675415039

Second set of inputs activating compound channel
original minus cython method q:  0.000000000000000 vel:  0.000000000000000 depth:  0.000000000000000
original minus cython method q:  0.000000000000000 vel:  0.000000000000000 depth:  0.000000000000000
original minus cython method q:  0.000000000000000 vel:  0.000000000000000 depth:  0.000000000000000
original minus cython method q:  0.000000000000000 vel:  0.000000000000000 depth:  0.000000000000000
original minus cython method q:  0.000000000000000 vel:  0.000000000000000 depth:  0.000000000000000
original minus cython method q:  0.000000000000000 vel:  0.000000000000000 depth:  0.000000000000000
original minus cython method q:  0.000000000000000 vel:  0.000000000000000 depth:  0.000000000000000
```
etc.

Now
Execute pytest -v  in that folder...
``` python
(base) [MC_singleSeg_singleTS]$ pytest -v
=================================================================== test session starts ====================================================================
platform linux -- Python 3.8.5, pytest-6.2.4, py-1.10.0, pluggy-0.13.1 -- bin/python
cachedir: .pytest_cache
rootdir: src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS
plugins: repeat-0.9.1
collected 3 items

test_MC_kernel.py::test_MC_kernel_q PASSED                                                                                                           [ 33%]
test_MC_kernel.py::test_MC_kernel_vel FAILED                                                                                                         [ 66%]
test_MC_kernel.py::test_MC_kernel_depth FAILED                                                                                                       [100%]

========================================================================= FAILURES =========================================================================
____________________________________________________________________ test_MC_kernel_vel ____________________________________________________________________

    def test_MC_kernel_vel():
    # TODO: Take advantage of ranges above to build a state-space
    # exploration of potential inputs and confirm it parity across all inputs

>       assert velc1 == velc2
E       assert 0.147629514336586 == 0.1286672204732895
E         +0.147629514336586
E         -0.1286672204732895

test_MC_kernel.py:81: AssertionError
___________________________________________________________________ test_MC_kernel_depth ___________________________________________________________________

    def test_MC_kernel_depth():
    # TODO: Take advantage of ranges above to build a state-space
    # exploration of potential inputs and confirm it parity across all inputs

>      assert depthc1 == depthc2
E      assert 85.61236572265625 == 39.90999984741211
E        +85.61236572265625
E        -39.90999984741211

test_MC_kernel.py:88: AssertionError
================================================================= short test summary info ==================================================================
FAILED test_MC_kernel.py::test_MC_kernel_vel - assert 0.147629514336586 == 0.1286672204732895
FAILED test_MC_kernel.py::test_MC_kernel_depth - assert 85.61236572265625 == 39.90999984741211
=============================================================== 2 failed, 1 passed in 1.32s ================================================================
(base) [MC_singleSeg_singleTS]$ pytest -v
=================================================================== test session starts ====================================================================
platform linux -- Python 3.8.5, pytest-6.2.4, py-1.10.0, pluggy-0.13.1 -- bin/python
cachedir: .pytest_cache
rootdir: src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS
plugins: repeat-0.9.1
collected 3 items

test_MC_kernel.py::test_MC_kernel_q PASSED                                                                                                           [ 33%]
test_MC_kernel.py::test_MC_kernel_vel PASSED                                                                                                         [ 66%]
test_MC_kernel.py::test_MC_kernel_depth PASSED                                                                                                       [100%]

==================================================================== 3 passed in 1.28s =====================================================================
```
Output appears to be volatile -- two identical pytest executions, different results. 

In the process of examining the errors, we added print statements to various portions of the calling stack. When the print statement is activated in the `src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/MCsingleSegStime_f2py_NOLOOP.f90`, the errors disappear, like what is reported [here ](https://stackoverflow.com/questions/27920043/adding-removing-a-print-statement-changes-variables) and [here](https://community.intel.com/t5/Intel-Fortran-Compiler/Adding-debugging-print-statements-changes-variable-values/td-p/740196).

### DIFF
--- a/src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/mc_sseg_stime_NOLOOP_demo.py
+++ b/src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/mc_sseg_stime_NOLOOP_demo.py
@@ -1,4 +1,6 @@
 import traceback
+from troute.routing.fast_reach import reach
+import test_suite_parameters
 
 debuglevel = 0
 COMPILE = True
@@ -157,6 +159,7 @@ def singlesegment(
 
     return rv[:3]
 
+
 def main():
     """
       No Inputs: 
@@ -303,24 +306,63 @@ def main():
     )
 
     print("\nSecond set of inputs activating compound channel")
-    dt = 60.0  # Time step
-    dx = 1800.0  # segment length
-    bw = 112.0  # Trapezoidal bottom width
-    tw = 248.0  # Channel top width (at bankfull)
-    twcc = 623.5999755859375  # Flood plain width
-    n_manning = 0.02800000086426735  # manning roughness of channel
-    n_manning_cc = 0.03136000037193298  # manning roughness of floodplain
-    cs = 0.42  # channel trapezoidal sideslope
-    s0 = 0.007999999690800905  # downstream segment bed slope
-    qlat = 40.0  # Lateral inflow in this time step
+    param_set = [
+        {
+            "precision": "single",
+            "dt": 60.0,  # Time step
+            "dx": 1800.0,  # segment length
+            "bw": 112.0,  # Trapezoidal bottom width
+            "tw": 248.0,  # Channel top width (at bankfull)
+            "twcc": 623.5999755859375,  # Flood plain width
+            "n_manning": 0.02800000086426735,  # manning roughness of channel
+            "n_manning_cc": 0.03136000037193298,  # manning roughness of floodplain
+            "cs": 0.42,  # channel trapezoidal sideslope
+            "s0": 0.007999999690800905,  # downstream segment bed slope
+            "qlat": 40.0,  # Lateral inflow in this time step
+            "qup": 45009,
+            "quc": 50098,
+            "qdp": 50014,
+            "depthp": 30,
+        },
+    ]
 
-    qup = 45009
-    quc = 50098
-    qdp = 50014
-    depthp = 30
+    qdc1, qdc2, velc1, velc2, depthc1, depthc2 = compare_methods(**param_set[0])
+
+    print(
+        "original minus cython method q: {0: 8.15f} vel: {1: 8.15f} depth: {2: 8.15f}".format(
+            qdc1 - qdc2, velc1 - velc2, depthc1 - depthc2
+        )
+    )
+
+    param_set = test_suite_parameters.test_suite_parameter_set
+    for p in param_set:
+        qdc1, qdc2, velc1, velc2, depthc1, depthc2 = compare_methods("single", *p)
+        print(
+            "original minus cython method q: {0: 8.15f} vel: {1: 8.15f} depth: {2: 8.15f}".format(
+                qdc1 - qdc2, velc1 - velc2, depthc1 - depthc2
+            )
+        )
+
+def compare_methods(
+    precision,
+    dt,
+    qup,
+    quc,
+    qdp,
+    qlat,
+    dx,
+    bw,
+    tw,
+    twcc,
+    n_manning,
+    n_manning_cc,
+    cs,
+    s0,
+    depthp,
+):
 
     # run M-C model
-    qdc, velc, depthc = singlesegment(
+    qdc1, velc1, depthc1 = singlesegment(
         # precision=precision,
         dt=dt,
         qup=qup,
@@ -338,13 +380,13 @@ def main():
         depthp=depthp,
     )
 
-    print(
-        "real {} precision computed via updated method q: {} vel: {} depth: {}".format(
-            precision, qdc, velc, depthc
-        )
-    )
+    # print(
+        # "real {} precision computed via updated method q: {} vel: {} depth: {}".format(
+            # precision, qdc1, velc1, depthc1
+        # )
+    # )
     # run M-C model
-    qdc, velc, depthc = singlesegment_wrf(
+    qdc_t, velc_t, depthc_t = singlesegment_wrf(
         # precision=precision,
         dt=dt,
         qup=qup,
@@ -361,11 +403,38 @@ def main():
         s0=s0,
         depthp=depthp,
     )
-    print(
-        "real {} precision computed via WRF-Hydro method q: {} vel: {} depth: {}".format(
-            precision, qdc, velc, depthc
-        )
+
+    # print(
+        # "real {} precision computed via WRF-Hydro method q: {} vel: {} depth: {}".format(
+            # precision, qdc_t, velc_t, depthc_t
+        # )
+    # )
+
+    # compare_func = reach.compute_reach_cython_kernel
+    compare_func = reach.compute_reach_kernel
+
+    # run M-C cython model
+    rv = compare_func(
+        # precision=precision,
+        dt,
+        qup,
+        quc,
+        qdp,
+        qlat,
+        dx,
+        bw,
+        tw,
+        twcc,
+        n_manning,
+        n_manning_cc,
+        cs,
+        s0,
+        0,
+        depthp,
     )
+    qdc2, velc2, depthc2 = rv["qdc"], rv["velc"], rv["depthc"]
+
+    return qdc1, qdc2, velc1, velc2, depthc1, depthc2
 
 
 if __name__ == "__main__":
--- a/src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/test_MC_kernel.py
+++ b/src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/test_MC_kernel.py
@@ -1,0 +1,90 @@
+import pytest
+import mc_sseg_stime_NOLOOP_demo as demo
+
+"""
+Statistics from the CONUS NWM 2.1 dataset
+            mean     std_dev      min        25%         50%          75%           max
+dx    1,947.776    1,965.625     1.000   558.000   1,549.000    2,631.000    95,714.000 
+bw        5.290        9.407     0.135     1.904       2.769        4.889       230.035 
+tw        8.817       15.678     0.225     3.173       4.615        8.148       383.392 
+twcc     26.450       47.033     0.674     9.518       13.845      24.443     1,150.175 
+n         0.058        0.003     0.040     0.060       0.060        0.060         0.060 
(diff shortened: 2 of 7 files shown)


jameshalgren:
@rcabell - this is the work I was referring to this afternoon. The conflicting results are troubling, but the magnitude of the error (when it appears) make me believe the error is not what causes the 'drift' we have observed in other cases. 

Ping also @awlostowski-noaa

awlostowski-noaa:
The test only fails for velocity and depth? That is strange. 

How did you find the parameter set that leads to failed tests? Are there others?

Let's say that depth values are occasionally erroneous. Then how does that error propagate to neighboring segments or time steps in the MC calculation? 

Also, totally bazar that you report errors going away when print statements are included.

rcabell:
Could this be a threading issue? It's possible that Python print statements have an implicit lock around them causing threads to (sort of) do a barrier-style resynchronize.

rcabell:
It's possible numpy is making some assumptions that the GIL is in place when called normally from Python code. If the cython functions in question are using `nogil ` you could try removing that temporarily to see if it makes a difference.

jameshalgren:
> The test only fails for velocity and depth? That is strange.
Yes. 

> How did you find the parameter set that leads to failed tests? Are there others?
The original was a fairly random parameter pick. I have now updated the function to use pseudo-random values, but a seed can be supplied so that the same test can be repeated. 

Every other test I performed in the first group (thousands of calculations, repeated many times) produced perfect matches. But now I'm realizing that was just by chance. The latest update shows that any number of evaluations (execute `python mc_sseg_stime_NOLOOP_demo.py`) will sometimes all fail (produce a delta result), though they usually all pass (produce 0 difference between functions). 

> Let's say that depth values are occasionally erroneous. Then how does that error propagate to neighboring segments or time steps in the MC calculation?
Good question. The depthp is part of the calculation of the qdc -- I'm digging in and will update.


> Also, totally bizarre that you report errors going away when print statements are included.
Yes, but not unheard of: https://stackoverflow.com/questions/33127940/result-changes-when-a-variable-is-used-in-print-statement-before-the-computation

jhrehanoaa (dismissed):

jhrehanoaa (approved):

groutr:
This quote from stackoverflow I think succinctly summarizes what @hellkite500 is pointing out

> Hmm, failing with -O3 while working with -O2 rather sounds like you are invoking UB and the compiler is completely free to do whatever it pleases. It's waaay easier to invoke UB compared to such a major feature of such a major and tested compiler to be broken.

We could try compiling with ubsan to check for undefined behavior. https://www.kernel.org/doc/html/latest/dev-tools/ubsan.html

hellkite500 (dismissed):
Setting optimization levels to -02 for now seems like the best way forward.  Will be interesting to see if we can pin down the root cause to some poorly defined behavior of gcc or gfortran using -03, but that might be a little tricky to do.

groutr:
@hellkite500 In my mind, the most likely place to look for UB is the interface between C and Fortran.

hellkite500:
> @hellkite500 In my mind, the most likely place to look for UB is the interface between C and Fortran.

Its definitely a place to start.  I just walked through the call stack to ensure that the cython code was calling the proper interface (`c_muskingcungenwm`) and that the fortran interface for `c_muskingcungenwm` used `bind( C )` and expected `c_float` input types.  All that looks good to me.

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/test_MC_kernel.py
```diff
@@ -0,0 +1,52 @@
+import pytest
+import mc_sseg_stime_NOLOOP_demo as demo
+from test_suite_parameters import generate_conus_MC_parameters
+
+
+input_tup = (
```
groutr, line 6: These tests should be restructured.

Consider:
```python
import pytest
import mc_sseg_stime_NOLOOP_demo as demo
from test_suite_parameters import generate_conus_MC_parameters

@pytest.fixture
def random_input_tuple():
	return generate_conus_MC_parameters(1, 16)

def test_MC_kernel(random_input_tuple):
	out = demo.compare_methods("single", *(next(params)))
	# print(out)
	qdc1, qdc2, velc1, velc2, depthc1, depthc2 = out
	assert (qdc1, velc1, depthc1) == (qdc2, velc2, depthc2)
```

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/test_suite_parameters.py
```diff
@@ -0,0 +1,94 @@
...
+    qlat = (1, 70000)
+    qup = (1, 70000)
+    quc = (1, 70000)
+    qdp = (1, 70000)
+    depthp = (1, 25)
+    dt = (5, 500)
+
+    np.random.seed(randomseed)
+
+    test_suite_parameter_set = zip_longest(
```
groutr, line 75: I think you should realize this into a sequence (either a list or tuple). What is the utility of returning an iterator?

jameshalgren:
I'm merging this PR with the test -- can we move further comments to #398?

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/test_MC_kernel.py
groutr: You might consider using hypothesis for generating test data.  https://hypothesis.readthedocs.io/en/latest/index.html

jameshalgren: @groutr: Is there a way to have that test result show the output for each test individually? It is very helpful (and potentially diagnostic) to know that the flow matches while the velocity and depth do not, for instance.

jameshalgren: Answering my own question --- `pytest -v`...

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/test_suite_parameters.py
jameshalgren: This usage in the non-automated test is the reason for generating a set in the first place: 
https://github.com/NOAA-OWP/t-route/blob/e90e84100354961b1de358fe77f7ed018827b410/src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/mc_sseg_stime_NOLOOP_demo.py#L347-L348

A list of tuples could work for that -- the overall output does need to be zipped together from the sequence of random inputs and it worked for the test without any transformation, so I left it that way.

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/mc_sseg_stime_NOLOOP_demo.py
```diff
@@ -1,4 +1,6 @@
 import traceback
+from troute.routing.fast_reach import reach
+from test_suite_parameters import generate_conus_MC_parameters
 
 debuglevel = 0
 COMPILE = True
```
jameshalgren, line 6: @groutr: Could this easily be turned into an input?

groutr: I don't see that being easy if not impossible. It would have to exist at a "meta-test" level.

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/test_MC_kernel.py
groutr: That's probably the best way. With pytest, I really try to avoid relying on a global state as it's not certain to always be consistent.

groutr: Another strategy is to use an addon for pytest, pytest-check (https://github.com/okken/pytest-check)

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/test_MC_kernel.py
```diff
@@ -0,0 +1,40 @@
+import pytest
+import mc_sseg_stime_NOLOOP_demo as demo
+from test_suite_parameters import generate_conus_MC_parameters
+
+
+@pytest.fixture
+def random_mc_input_tuple():
+    return generate_conus_MC_parameters(1,16)
+
+"""
```
groutr, line 10: Should this be in the docstring for demo.compare_methods?

groutr (dismissed):

src/fortran_routing/mc_pylink_v00/MC_singleSeg_singleTS/test_suite_parameters.py
jameshalgren: That would make sense... for a shorter input. We are creating thousands of outputs for the demo test. Does a list still make sense in that case?

PR #394, opened 2021-09-01
Looping da #1

jhrehanoaa:
Contains looping of da subsets through the same methods as developed for q_lats. Instead of changing the functions that ingest da data to match run_sets[], I just kept data_assimilation_parameters the same to be passed through everything.

The main change comes from assigning:

data_assimilation_parameters['data_assimilation_filter'] = run_sets_da[X]['data_assimilation_subset']

in different instances for the first and following iterations.

That way data_assimilation_parameters['data_assimilation_filter'] changes and is passed as the current subset each iteration.


jhrehanoaa:
Replaced in dynamic da looping PR. Combines both block calls and dynamic date range block creation.

PR #395, opened 2021-09-01
Da loop dynamic #2

jhrehanoaa:
This PR builds upon the previous da looping PR. The other PR contains the operational components to run data assimilation in a block test format the same way qlats were handled. In this new PR we are replacing the blocks with a start and end date with a new block size. The code then can derive the block sizing of da files and build the same blocks that were manually entered in the yaml in an automatic fashion. This will streamline large testing of much larger time spans with looping. Mostly operates within .io file to run the function def build_da_date_range(data_assimilation_parameters): in order to achieve this goal. 

Should be integrated after da_looping PR. Need to add preservation of block input approach AND dynamic block building approach for da input files.

### DIFF
--- a/src/nwm_routing/src/nwm_routing/preprocess.py
+++ b/src/nwm_routing/src/nwm_routing/preprocess.py
@@ -50,7 +50,7 @@ def nwm_network_preprocess(
     # waterbodies_values = supernetwork_values[12]
     # waterbodies_segments = supernetwork_values[13]
     # connections_tailwaters = supernetwork_values[4]
-
+    waterbody_type_specified = False
     if break_network_at_waterbodies:
         # Read waterbody parameters
         waterbodies_df = nhd_io.read_waterbody_df(
@@ -73,8 +73,6 @@ def nwm_network_preprocess(
 
         wbtype="level_pool"
         wb_params_level_pool = waterbody_parameters.get(wbtype, defaultdict(list))  # TODO: Convert these to `get` statments
-
-        waterbody_type_specified = False
 
         # NOTE: What are we accomplishing with this logic here?
         if wb_params_hybrid_and_rfc["reservoir_persistence_usgs"] \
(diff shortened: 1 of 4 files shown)


test/input/yaml/CustomInput_loop_da.yaml
```diff
@@ -0,0 +1,155 @@
+---
+#initial input parameters
+log_parameters:
+    verbose: true  # verbose output (leave blank for quiet output.)
+    showtiming: true  # set the showtiming (omit flag for no timing information.)
+    debuglevel: 1  # set the debuglevel for additional console output.
+network_topology_parameters:
+    supernetwork_parameters:
+        title_string: "Florence_FullRes with Nudging in Forecast Mode"
+        geo_file_path: "/glade/scratch/jamesmcc/tmp_florence_run_nudging/forecasts/cast_2018091800/NWM/DOMAIN/Route_Link.nc"
```
jameshalgren, line 10: We can't have system file paths -- everything has to reference the repository, even if it is a symlinked folder that is broken on another system.

src/nwm_routing/src/nwm_routing/preprocess.py
```diff
@@ -190,17 +193,13 @@ def nwm_initial_warmstate_preprocess(
...
     if verbose:
         print("channel initial states complete")
     if showtiming:
         print("... in %s seconds." % (time.time() - start_time))
         start_time = time.time()
 
-    # TODO: Does this need to live outside the if-block for waterbodies above? If not, let's move it up there to keep things together.
-    waterbodies_df = pd.merge(
-        waterbodies_df, waterbodies_initial_states_df, on="lake_id"
-    )
```
jameshalgren, line 203: - [ ] need to check whether this is relevant to the PR -- if not, we can wrap it in a quick separate PR.

jameshalgren:
- [ ] add warm state capability
At some level, we'll need to do more in here to handle the warm states. For instance, the `lastobs_df` returned [here](https://github.com/NOAA-OWP/t-route/blob/3a259be2950d144538eaa6059c0e5b7080175aee/src/nwm_routing/src/nwm_routing/__main__.py#L883) is going to need to be specified in the input file, and we need to verify that the function is doing what we expect. And the [`q0` warm-state](https://github.com/NOAA-OWP/t-route/blob/3a259be2950d144538eaa6059c0e5b7080175aee/src/nwm_routing/src/nwm_routing/__main__.py#L997) will need a corresponding `lastobs` value passed out of the compute process and made available to hand back to the next portion of the looped execution.

jameshalgren:
-[ ] ensure functional wrapping
All the algorithmic details in the `main_v03` function need to be wrapped in a sub-function or be one-liners in order to be compatible with eventual use with the asynchronous I/O capability.

PR #396, opened 2021-09-01
Qlat looping dynamic #3

jhrehanoaa:
This is the third PR involving looping that will follow after the integration of the previous two. It mimics the functionality of the dynamic da block creation and creates the same exact structure used in the yaml file previously in an automatic way. NTS is also created in this looping. This way nothing in the compute functionality changes and instead we can just create the blocked test segments of qlat files quickly at the start of the model process and save it for later. 

def build_qlat_date_range(forcing_parameters):  and def build_qlat_sets(forcing_parameters,dates): within the .io file achieves most of this. Need to preserve block approach or dynamic approach for qlat input file blocks.

### DIFF
--- a/src/nwm_routing/src/nwm_routing/__main__.py
+++ b/src/nwm_routing/src/nwm_routing/__main__.py
@@ -898,6 +898,7 @@ def main_v03(argv):
     run_sets = forcing_parameters.get("qlat_forcing_sets", False)
     run_sets_da = data_assimilation_parameters.get("data_assimilation_sets", False)
     if run_sets_da:
+        nhd_io.build_da_sets(data_assimilation_parameters,0)
         data_assimilation_parameters['data_assimilation_filter'] = run_sets_da[0]['data_assimilation_subset']
     # TODO: Data Assimilation will be something like the parity block
     # if DA:
--- a/src/nwm_routing/src/nwm_routing/__main__.py
+++ b/src/nwm_routing/src/nwm_routing/__main__.py
@@ -897,9 +897,7 @@ def main_v03(argv):
     # TODO: Make this more flexible.
     # run_sets_qlats = forcing_parameters.get("qlat_files", False)
 
-    forcing_parameters = nhd_io.build_qlat_date_range(forcing_parameters)
-    
-    run_sets = forcing_parameters['qlat_forcing_sets']["qlat_files"]
+    run_sets = nhd_io.build_qlat_date_range(forcing_parameters)
     run_sets_da = data_assimilation_parameters.get("data_assimilation_sets", False)
     run_size_da = run_sets_da['data_assimilation_run_block_size']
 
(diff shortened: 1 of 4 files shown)


src/python_framework_v02/troute/nhd_io.py
```diff
@@ -949,3 +952,51 @@ def build_coastal_ncdf_dataframe(coastal_ncdf):
...
+    date_time_str = data_assimilation_end[:13]
+    date_time_obj_end = datetime.strptime(date_time_str, "%Y-%m-%d_%H")
+
+    dates = []
+    # for j in pd.date_range(date_time_obj_start, date_time_obj_end + timedelta(1), freq="5min"):
+    for j in pd.date_range(date_time_obj_start, date_time_obj_end, freq="1H"):
+        dates.append(j.strftime("%Y-%m-%d_%H:%M:00")+str(file_tail))
+    return dates
+
+def build_qlat_date_range(forcing_parameters):
```
jhrehanoaa, line 974: @BrianAvant-NOAA This is the funtion that builds up the blocks of qlats from a start date and end date. This should be integrated by next week.

src/nwm_routing/src/nwm_routing/__main__.py
```diff
@@ -896,8 +895,23 @@ def main_v03(argv):
     # The inputs below assume a very pedantic setup
     # with each run set explicitly defined, so...
     # TODO: Make this more flexible.
+
     run_sets = forcing_parameters.get("qlat_forcing_sets", False)
+    
+    if type(run_sets) == list:
+        pass
+    else:
+        run_sets = nhd_io.build_qlat_date_range(forcing_parameters)
```
jhrehanoaa, line 904: @BrianAvant-NOAA Here is the check that determines if you want to pass the qlats in using the old block format or the dynamic start->end date setup.

test/input/yaml/CustomInput_loop_da.yaml
```diff
@@ -0,0 +1,159 @@
...
+        # means the least ambiguity in how we intend to set things up.
+        # a more sophisticated algorithm could accept a glob list
+        # and break that up according to some `qlat_forcing_configuration`
+        # but for now, we'll use this.
+        qlat_forcing_sets:
+            qlat_file_range:
+                qlat_start_file: "201809180100.CHRTOUT_DOMAIN1"
+                qlat_end_file: "201809180600.CHRTOUT_DOMAIN1"
+        qlat_run_block_size: 4
+            # - qlat_files:
```
jhrehanoaa, line 97: @BrianAvant-NOAA you can uncomment the blocks to run code that way, but over longer periods of time, it will make sense to do it dynamically with the file range in line 94.

jhrehanoaa:
replaced by another PR with dynamic parity

PR #397, opened 2021-09-01
Async looping merged #4

jhrehanoaa:
Carried over the necessary changes from the async branch to match those with the looping capabilities. Will need to create the comparison or explanation for the repository next that walks through what is taking place. Made all the necessary changes in main.py file.


jameshalgren:
We'll need to rework this to be an optional capability that can be triggered as async or synchronous looping.

Issue #391, opened 2021-08-24
Waterbody splitting causes massive performance impact 


jameshalgren:
@rcabell identified an issue that when the `split_network_at_waterbodies` flag is set to true, the algorithm gets stuck  inside the `organize_independent_networks` function for a very large scale simulation. Removing the flag, the process completes in ~20 seconds; with the flag set, it ran for over an hour or never completed. The traceback from interrupting the loop shows the following:

```
Traceback (most recent call last):
  File "t-route/src/nwm_routing/src/nwm_routing/__main__.py", line 1030, in <module>
    main_v02(v_args[1])
  File "t-route/src/nwm_routing/src/nwm_routing/__main__.py", line 397, in main_v02
    independent_networks, reaches_bytw, rconn = nnu.organize_independent_networks(
  File "t-route/src/python_framework_v02/troute/nhd_network_utilities_v02.py", line 513, in organize_independent_networks
    reaches_bytw[tw] = nhd_network.dfs_decomposition(net, path_func)
  File "t-route/src/python_framework_v02/troute/nhd_network.py", line 393, in dfs_decomposition
    if path_func(path, n):
  File "t-route/src/python_framework_v02/troute/nhd_network.py", line 206, in split_at_waterbodies_and_junctions
    if (path[-1] in waterbody_nodes) ^ (node in waterbody_nodes):
```

PR #400, opened 2021-09-02
change list of waterbodies to sets in split_at_waterbodies_and_juncti…

awlostowski-noaa:
This PR addresses the issue of very long evaluation times needed to run `nnu.organize_independent_networks` at CONUS scale with splitting at waterbodies (#391). This function was taking hours to run, and now takes seconds. The root cause of the issue was that `nhd_network.split_at_waterbodies_and_junctions` was searching through LISTS of waterbodies, which, when large, are burdensome to search. The fix implemented is to make sure that SETS of waterbodies are passed to this function. 

Before this fix, the  `nnu.organize_independent_networks` took over an hour to evaluate. Now, it takes 21 seconds.

### DIFF
--- a/src/nwm_routing/src/nwm_routing/__main__.py
+++ b/src/nwm_routing/src/nwm_routing/__main__.py
@@ -313,6 +313,9 @@ def main_v02(argv):
     break_network_at_waterbodies = run_parameters.get(
         "break_network_at_waterbodies", False
     )
+    break_network_at_gages = supernetwork_parameters.get(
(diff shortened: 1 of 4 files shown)


jameshalgren:
@rcabell -- this should resolve the hangup you were observing with the CONUS test.

rcabell:
Thanks! — I’ll try this out and let you know the results

jameshalgren (approved):

PR #404, opened 2021-09-06
bugfix: gages in offnetwork upstream list blows up by-subnetwork parallel simulations

awlostowski-noaa:
by-subnetwork parallel simulations were breaking with the Florence_Benchmark_da simulations. After some digging, I found that there was a gage link in the offnetwork upstreams list. This caused `da_positions_list_byreach` to be one item shorter than `da_positions_list_byseg`, which blew up the simulation with an indexing error in `mc_reach.compute_network_structured`. 

To fix this, I pass the `offnetwork_upstreams` list to `_prep_da_dataframes` (default  `offnetwork_upstreams = []`), in addition to `param_df_sub.index`. Then I create a new list called, `subnet_segs`, which contains ONLY the segments in the subnetwork. This list is searched for gage segments. The `param_df_sub_idx` list is handled with `get_indexer()` such that the `param_df_sub` array is correctly indexed in downstream operations. 

I tested this scheme with `Florence_Benchmark_da.yaml`, all gage segments, all parallel schemes. The same results at gage locations were observed across all parallel schemes, which proves that the fix is robust.

### DIFF
--- a/src/python_routing_v02/troute/routing/compute.py
+++ b/src/python_routing_v02/troute/routing/compute.py
@@ -334,7 +334,7 @@ def compute_nhd_routing_v02(
                         param_df_sub.index.tolist() + lake_segs
                     ).sort_index()
 
-                    usgs_df_sub, lastobs_df_sub, da_positions_list_byseg = _prep_da_dataframes(usgs_df, lastobs_df, param_df_sub.index)
+                    usgs_df_sub, lastobs_df_sub, da_positions_list_byseg = _prep_da_dataframes(usgs_df, lastobs_df, param_df_sub.index.symmetric_difference(offnetwork_upstreams))
                     da_positions_list_byreach, da_positions_list_bygage = _prep_da_positions_byreach(subn_reach_list, lastobs_df_sub.index)
 
                     qlat_sub = qlat_sub.reindex(param_df_sub.index)
@@ -547,7 +547,7 @@ def compute_nhd_routing_v02(
                         param_df_sub.index.tolist() + lake_segs
                     ).sort_index()
 
-                    usgs_df_sub, lastobs_df_sub, da_positions_list_byseg = _prep_da_dataframes(usgs_df, lastobs_df, param_df_sub.index)
+                    usgs_df_sub, lastobs_df_sub, da_positions_list_byseg = _prep_da_dataframes(usgs_df, lastobs_df, param_df_sub.index.symmetric_difference(offnetwork_upstreams))
                     da_positions_list_byreach, da_positions_list_bygage = _prep_da_positions_byreach(subn_reach_list, lastobs_df_sub.index)
 
                     qlat_sub = qlat_sub.reindex(param_df_sub.index)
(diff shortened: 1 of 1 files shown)


awlostowski-noaa:
ping @jameshalgren 
I fixed the erroneous indexing.

src/python_routing_v02/troute/routing/compute.py
```diff
@@ -71,18 +72,20 @@ def _prep_da_dataframes(
...
     # NOTE: Uncomment to easily test no observations...
     # usgs_df = pd.DataFrame()
     if not usgs_df.empty and not lastobs_df.empty:
         # index values for last obs are not correct, but line up correctly with usgs values. Switched
-
-        lastobs_segs = list(lastobs_df.index.intersection(param_df_sub_idx))
+        lastobs_segs = list(lastobs_df.index.intersection(subnet_segs))
         lastobs_df_sub = lastobs_df.loc[lastobs_segs]
-        usgs_segs = list(usgs_df.index.intersection(param_df_sub_idx))
+        usgs_segs = list(usgs_df.index.intersection(subnet_segs))
```
awlostowski-noaa, line 84: Find gage segments for DA by intersection of `usgs_df.index` with `subnet_segs`, which DOES NOT contain any segments in `offnetwork_upstreams`

src/python_routing_v02/troute/routing/compute.py
```diff
@@ -71,18 +72,20 @@ def _prep_da_dataframes(
...
         da_positions_list_byseg = param_df_sub_idx.get_indexer(usgs_segs)
```
awlostowski-noaa, line 85: This line remains unchanged, ensuring that the `da_positions_list_byseg` is correctly indexing `param_df_sub_idx`

src/python_routing_v02/troute/routing/compute.py
```diff
@@ -71,18 +72,20 @@ def _prep_da_dataframes(
     time series is as long as the simulation.
 
     """
+    # segments in the subnetwork ONLY, no offnetwork upstreams included
+    subnet_segs = param_df_sub_idx.symmetric_difference(offnetwork_upstreams)
```
jameshalgren, line 76: I think we would want `difference`.

awlostowski-noaa: rgr. changed.

jameshalgren:
One final comment -- suggest using set, not list. Try this: 
``` py
diff --git a/src/python_routing_v02/troute/routing/compute.py b/src/python_routing_v02/troute/routing/compute.py
index 3d5a3ce5..c72bfd7d 100644
--- a/src/python_routing_v02/troute/routing/compute.py
+++ b/src/python_routing_v02/troute/routing/compute.py
@@ -52,7 +52,7 @@ def _prep_da_dataframes(
     usgs_df,
     lastobs_df,
     param_df_sub_idx,
-    offnetwork_upstreams=[],
+    exclude_segments=None,
     ):
     """
     Produce, based on the segments in the param_df_sub_idx (which is a subset
@@ -60,6 +60,8 @@ def _prep_da_dataframes(
     a subset of the relevant usgs gage observation time series
     and the relevant last-valid gage observation from any
     prior model execution.
+    Gages in `exclude_segments` will not be included in the
+    returned indexes.

     Cases to consider:
     USGS_DF, LAST_OBS
@@ -73,8 +75,9 @@ def _prep_da_dataframes(

     """
     # segments in the subnetwork ONLY, no offnetwork upstreams included
-    subnet_segs = param_df_sub_idx.difference(offnetwork_upstreams)
-
+    subnet_segs = param_df_sub_idx
+    if exclude_segments:
+        subnet_segs = subnet_segs.difference(set(exclude_segments))
     # NOTE: Uncomment to easily test no observations...
     # usgs_df = pd.DataFrame()
     if not usgs_df.empty and not lastobs_df.empty:
```

jameshalgren (approved):
@awlostowski-noaa -- thanks.

PR #407, opened 2021-09-07
Allow flexible output folder for WRF-Hydro output files

jameshalgren:
The WRF-Hydro-style output currently relies on an existing output folder, which is scanned, and then a duplicate set of output files are generated (CHRTOUT or HYDRO_RST) with the t-route streamflow values populating the results. 

This PR allows for that write to occur in a different directory than the source files which are scanned.


awlostowski-noaa (approved):

jhrehanoaa:
I can confirm the Inland_config.yaml file works on this branch.

PR #412, opened 2021-09-08
Allow flexible output folder

jameshalgren:
Bugfix PR for #407 -- the attaches the now-default folders for the output of the CustomInput.yaml test.


jameshalgren:
admin pushed -- these directories were dropped from a commit and were breaking the CustomInput.yaml test.

PR #413, opened 2021-09-08
Allow flexible output folder loop

jameshalgren:
This is the part of updating the looping execution capability. In particular, this creates more consistent alignment of the output for CHRTOUT and HYDRO_RST files during looped execution. There may be additional needs regarding the development of regular warm-state output files.

Important change: 
This code now provides a t0 parameter computed at the beginning of each run. 

The code can be demonstrated with `python -m nwm_routing -V3 -f ../../test/input/yaml/CustomInput_loop.yaml`


src/nwm_routing/src/nwm_routing/__main__.py
```diff
@@ -998,17 +999,20 @@ def main_v03(argv):
...
         nwm_output_generator(
+            run,
             run_results,
             supernetwork_parameters,
             output_parameters,
             parity_parameters,
+            restart_parameters,
             parity_sets[run_set_iterator] if parity_parameters else {},
-            nts,
+            qts_subdivisions,
```
jameshalgren, line 1009: Does this belong inside of the `run` dict?

src/nwm_routing/src/nwm_routing/__main__.py
```diff
@@ -998,17 +999,20 @@ def main_v03(argv):
...
             compute_parameters.get("return_courant", False),
             showtiming,
             verbose,
             debuglevel,
         )
+        t0 = t0 + timedelta(seconds = dt * nts)
```
jameshalgren, line 1015: Print this to verify that it is incrementing properly.

src/nwm_routing/src/nwm_routing/preprocess.py
```diff
@@ -203,9 +221,10 @@ def nwm_initial_warmstate_preprocess(
     )
 
     last_obs_file = restart_parameters.get("wrf_hydro_last_obs_file", None)
+    # TODO: Split apart the da input functions to get last_obs here.
```
jameshalgren, line 224: - [ ] make Issue to capture this TODO.

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -904,13 +905,19 @@ def write_channel_restart_to_wrf_hydro(
     -------
     """
     # create t-route simulation timestamp array
-    with xr.open_dataset(channel_initial_states_file) as ds:
-        t0 = ds.Restart_Time.replace("_", " ")
+    # TODO: Replace this with a more general function to identify
```
jameshalgren, line 908: - [ ] Make issue to capture this TODO

src/nwm_routing/src/nwm_routing/preprocess.py
```diff
@@ -1,11 +1,26 @@
...
 
+def get_t0_str_from_restart_parameters(
+    restart_parameters,
+):
+    if restart_parameters.get("wrf_hydro_channel_restart_file",None):
+        channel_initial_states_file = restart_parameters["wrf_hydro_channel_restart_file"]
+        with xr.open_dataset(channel_initial_states_file) as ds:
+            t0_str = ds.Restart_Time.replace("_", " ")
+    else:
+        t0_str = "2015-08-16 00:00:00"
```
awlostowski-noaa, line 19: If we don't have a restart file, I suppose we are just cold starting from an arbitrary t0 (this date). Do we have the capability to read restarts from csv? If so, do those files contain any sort of start information?

awlostowski-noaa (commented):
What happens if the t0 of the restart file is not the timestep prior to the first CHRTOUT file? Should we throw an error? The synchrony of the restart and CHRTOUT files is an important or ensuring that the simulations are temporally coherent.

src/nwm_routing/src/nwm_routing/preprocess.py
jameshalgren: The restart from CSV capability permits but does not require date-centric headers for the CSV columns. The two places that the t0 is used directly are 1: for the Reservoir simulation; 2: for dealing with the restart file write out. 

We could add a parameter to the yaml file to allow the user to directly specify a t0. That might be used to provide a value to accompany a csv file lacking the information otherwise. It might also, either intentionally or accidentally, override the value in the file. 

My recommendation is to avoid over complicating the input handling -- We are designing most specifically for the operational case, where a t0 will always be available and known. If we require the use to input that at the front, then it makes it easy for a later effort to create a wrapper that provides it by more sophisticated heuristics.

jameshalgren:
- [ ] backport t0 to v02 execution

src/nwm_routing/src/nwm_routing/preprocess.py
jameshalgren: Happening now as part of the da looping integration.

src/nwm_routing/src/nwm_routing/__main__.py
jameshalgren: Verified.

src/nwm_routing/src/nwm_routing/__main__.py
jameshalgren: Working on this as part of da looping integration.

src/python_framework_v02/troute/nhd_io.py
jameshalgren: This is now reading a t0 from the input, which is pulled from a general function.

src/nwm_routing/src/nwm_routing/preprocess.py
jameshalgren: Setting this out as issue #427

awlostowski-noaa (approved):

PR #420, opened 2021-09-13
[WIP] CNT model - with conflicts

awlostowski-noaa:
Introducing the diffusive wave CNT code. 

The `diffusive_cnt` is a new `compute_kernel` option that executes the "CNT" formulation of the diffusive wave code. The contents of this pull request integrate the CNT kernel into the framework with all necessary Fortran-cython-python workflows. 

## known issues
- code is unstable when dt < 450 seconds
- Depth, celerity, and diffusivity are arbitrarily initialized in the Fortran kernel. Only initial flow conditions are accepted.


kumdonoaa (approved):
still need to figure out handling of time step variables for forcing data, output publishing, and internal simulation time step.

Issue #427, opened 2021-09-14
Handle (more elegantly) restart date


jameshalgren:
The original questions...
>If we don't have a restart file, I suppose we are just cold starting from an arbitrary t0 (this date). Do we have the capability to read restarts from csv? If so, do those files contain any sort of start information?
and
> What happens if the t0 of the restart file is not the timestep prior to the first CHRTOUT file? Should we throw an error? The synchrony of the restart and CHRTOUT files is an important or ensuring that the simulations are temporally coherent.

... both refer to this line, where the fallback t0 date is set to the National Water Model birthday. 
https://github.com/jameshalgren/t-route/blob/f95308ff6fb65204a511e2547e0f28c846047efd/src/nwm_routing/src/nwm_routing/preprocess.py#L161

We might have no restart or a restart in a csv file. And with either case, there may not be coherence between the restart file and the forcing inputs. None of those cases are particularly well handled by the current code. 

The tasks for this issue are generally:
- [ ] Handle no restart date
- [ ] Handle restart in csv
- [ ] Warn or otherwise handle non-corresponding restart and forcing.

_Originally posted by @awlostowski-noaa in https://github.com/NOAA-OWP/t-route/pull/413#discussion_r707304106_

jameshalgren:
The restart from CSV capability permits but does not require date-centric headers for the CSV columns. The two places that the t0 is used directly are 1: for the Reservoir simulation; 2: for dealing with the restart file write out.

We could add a parameter to the yaml file to allow the user to directly specify a t0. That might be used to provide a value to accompany a csv file lacking the information otherwise. It might also, either intentionally or accidentally, override the value in the file.

My recommendation is to avoid over complicating the input handling -- We are designing most specifically for the operational case, where a t0 will always be available and known. If we require the use to input that at the front, then it makes it easy for a later effort to create a wrapper that provides it by more sophisticated heuristics.

PR #429, opened 2021-09-14
Long term looping all elements

jameshalgren:
### TL; DR -- Full V3 "Looping" capability implemented
This PR implements the full input interface functionality of the NWM routing module simulation elements with
- [x] Long-term simulation for large domains (called "looping" or "V3" execution)
- [x] Reservoir simulation capability (newly integrated into V3)
- [x] Data assimilation capability (newly integrated into V3)
Follow-on PRs will address the following
- [ ] Handling non-numeric gage ids (from Canadian and other non-USGS databases)
- [ ] Trigger LastObs file write
- [ ] Incorporate asynchronous file I/O at the looping level

***Note***: The V3 designation is only loosely tied to the NWM v3.0 version number for which this capability is targeted; Semantically, this essentially represents a new version of the model, as the input version is now completely different and significant new functionality is introduced. 

### Major work
* Update front-end interface
* Split LastObs and TimeSlice preprocessing to recognize relation to initial states vs. continual forcing
* Add additional return from compute kernel with LastObs updated through simulated series
* Use returned lastobs with other warm states
* Adjust logic for TimeSlice read to accommodate repeated calls in consecutive loops
* Catch bugfixes in the less-tested looping execution pathway
* Prepare a test case utilizing all input capacity (thanks @jmccreight!)
* ...
#### Changes to front end
V2: Existing workflow to execute routing calculations and control outputs
```python -m nwm_routing -V2 -f CustomInput.yaml```
The `.yaml` file architecture you are currently used to seeing
V3 (“looping”): New workflow permitting sequentially connected simulations 
```python -m nwm_routing -V3 -f CustomInput_loop.yaml```
A new `.yaml` configuration where user specifies forcing files for each simulation batch in the sequence (next slides)
#### Modified DA Functions
Functions to handle reading of data assimilation timeslice files and lastobs values were split to permit alternate handling in the different parts of looped execution. 
The updated functions are here
https://github.com/jameshalgren/t-route/blob/a4b1d2084c2ffe7e8104ba0222fdd5a317369afe/src/python_framework_v02/troute/nhd_network_utilities_v02.py#L629-L691
With the new calls split here: 
https://github.com/NOAA-OWP/t-route/blob/a4b1d2084c2ffe7e8104ba0222fdd5a317369afe/src/nwm_routing/src/nwm_routing/preprocess.py#L328
... and here:
https://github.com/NOAA-OWP/t-route/blob/a4b1d2084c2ffe7e8104ba0222fdd5a317369afe/src/nwm_routing/src/nwm_routing/preprocess.py#L232-L234

... and the unmodified V2 version call here:
https://github.com/jameshalgren/t-route/blob/a4b1d2084c2ffe7e8104ba0222fdd5a317369afe/src/nwm_routing/src/nwm_routing/__main__.py#L576-L578
(notice how the prior version uses a combined call.)

#### Return LastObs from Compute Kernel
With this simple change, the `compute_network_structured` compute kernel now returns an additional element in the return array -- a tuple of arrays of the form `([gage ids], [timesincelastobs], [lastobsvalues])`:
https://github.com/NOAA-OWP/t-route/blob/a4b1d2084c2ffe7e8104ba0222fdd5a317369afe/src/python_routing_v02/troute/routing/fast_reach/mc_reach.pyx#L1416 
This change affects the kernel which will be used for operational execution; an issue has been logged to include the same functionality with the other compute kernels. A placeholder value (just a '0') is used for the return of diagnostic outputs such as courant number, which are returned by other kernels but are not yet possible to return from the operational kernel (see #362 and #380). 
#### LastObs Warmstate
The LastObs arrays returned from the last timestep of the calculation are refomatted into the dataframe expected as input with this new function:
https://github.com/jameshalgren/t-route/blob/a4b1d2084c2ffe7e8104ba0222fdd5a317369afe/src/nwm_routing/src/nwm_routing/__main__.py#L956-L971
The call to this new function is grouped with the other calls to prepare warmstates to 'bootstrap' the execution of the next loop here: 
https://github.com/jameshalgren/t-route/blob/a4b1d2084c2ffe7e8104ba0222fdd5a317369afe/src/nwm_routing/src/nwm_routing/__main__.py#L1122-L1126
#### Bugfixes
See principally commits in 
https://github.com/NOAA-OWP/t-route/pull/429/commits/e640e8cdf0b1bcd4d1c15d7b16e50327b21a13bb
https://github.com/NOAA-OWP/t-route/pull/429/commits/69f6f4aa4883c43a333052009f1d4ec6cf634cdb
#### Test case
A test case with five minute output, gage data assimilation, and reservoir simulation was prepared for testing on the Cheyenne computing infrastructure. 

### Testing/Screenshots
These screenshots show output from our "parity check", which is a simple comparison of WRF-Hydro simulated output and the same timeseries from the t-route model. A more comprehensive statistical comparison tool is in progress (and will help a lot!!) For now, we see that ***the V3 update retains perfect fidelity with the earlier V2 capability<sup>*</sup>***, and produces no change in the fidelity with respect to the WRF-Hydro output. 

<sup>*</sup>When there is a gap in the assimilation time series and that gap splits over the looping breaks, the interpolation of the time series is interrupted. The solution is to pad the sequence of assimilation files to cover over the gap. Commit 
9a0b7a8 adds capability to trim the padded time series to the relevant portion for the particular portion of the loop. 
#### Gage site 8778363
Tests on the Cheyenne HPC system confirm that ***V2 and V3 produce identical results***: 
V3 first five timesteps parity check at gage site 8778363:
image
Last five (split up because the looping breaks the simulation into several chunks):
image

V2 first and last five at same site, with summary statistics:
image

#### Inflow to Lake Butner
V3 at 8777125, inflow to Lake Butner
image
image
V2, same
image

##### Reservoir Outflow, DA OFF
V3 at 166737579, Falls Lake outflow first and last five minutes __Data Assimilation OFF__. Note the deviation at the end of the time series due to lake of assimilation in the upstream network.
image
image

V2, same
image
##### Reservoir Outflow, DA ON
There is a slight difference between these series. See commit 9a0b7a8 and comments below for correction and explanation. 
V3, same __DA ON__
image
image
V2, same
image
##### Falls Lake Inflow, DA ON
V3 at 8782449, Falls Lake Inflow __DA ON__
image
image
V2, same
image

#### DA Smoothing
While not the primary purpose of this development, with the five-minute output files to compare to with this test, it is easy to see the slight difference in the treatment of interpolation with the data assimilation scheme. T-route takes advantage of a linear interpolation to smooth the DA values between 15 minute observations vs. a more step-function approach in the WRF-Hydro DA algorithm. The following output shows this fairly clearly -- the difference values oscillate back and forth as the linear interpolation moves across the step-function value:
image

### DIFF
--- a/src/nwm_routing/src/nwm_routing/__main__.py
+++ b/src/nwm_routing/src/nwm_routing/__main__.py
@@ -1057,6 +1057,7 @@ def main_v03(argv):
         break_network_at_waterbodies,
         param_df.index,
         lastobs_df.index,
+        t0,
         showtiming,
         verbose,
         debuglevel,
@@ -1064,7 +1065,7 @@ def main_v03(argv):
 
     for run_set_iterator, run in enumerate(run_sets):
 
-        run["t0"] = t0
+        t0 = run.get("t0")
         dt = run.get("dt")
         nts = run.get("nts")
 
@@ -1114,6 +1115,7 @@ def main_v03(argv):
                 break_network_at_waterbodies,
                 param_df.index,
                 lastobs_df.index,
+                t0 + timedelta(seconds = dt * nts),
                 showtiming,
                 verbose,
                 debuglevel,
@@ -1141,7 +1143,6 @@ def main_v03(argv):
             verbose,
             debuglevel,
         )
-        t0 = t0 + timedelta(seconds = dt * nts)
 
     # nwm_final_output_generator()
 
--- a/src/nwm_routing/src/nwm_routing/__main__.py
+++ b/src/nwm_routing/src/nwm_routing/__main__.py
@@ -1121,8 +1121,10 @@ def main_v03(argv):
 
             q0 = new_nwm_q0(run_results)
 
-            lastobs_df = new_lastobs(run_results, dt * nts)
+            if data_assimilation_parameters:
+                lastobs_df = new_lastobs(run_results, dt * nts)
 
+            # TODO: Confirm this works with Waterbodies turned off
             waterbodies_df = get_waterbody_water_elevation(waterbodies_df, q0)
 
         nwm_output_generator(
--- a/src/nwm_routing/src/nwm_routing/__main__.py
+++ b/src/nwm_routing/src/nwm_routing/__main__.py
@@ -3,6 +3,7 @@ import time
 from datetime import datetime, timedelta
 from collections import defaultdict
 from pathlib import Path
+import numpy as np
 import pandas as pd
 
 ## network and reach utilities
@@ -952,9 +953,22 @@ def get_waterbody_water_elevation(waterbodies_df, q0):
     return waterbodies_df
 
 
-def new_lastobs(run_results):
-    pass
-
+def new_lastobs(run_results, time_increment):
+    df = pd.concat(
+        [
+            pd.DataFrame(
+                # TODO: Add time_increment (or subtract?) from time_since_lastobs
+                np.array([rr[3][1],rr[3][2]]).T,
+                index=rr[3][0],
+                columns=["time_since_lastobs", "lastobs_discharge"]
+            )
+            for rr in run_results
+            if not rr[3][0].size == 0
+        ],
+        copy=False,
+    )
+    df["time_since_lastobs"] = df["time_since_lastobs"] - time_increment
+    return df
 
 def main_v03(argv):
     args = _handle_args_v03(argv)
@@ -1104,9 +1118,9 @@ def main_v03(argv):
                 debuglevel,
             )
 
-            # q0 = run_results
             q0 = new_nwm_q0(run_results)
-            # lastobs_df = new_lastobs(run_results)
+
+            lastobs_df = new_lastobs(run_results, dt * nts)
 
             waterbodies_df = get_waterbody_water_elevation(waterbodies_df, q0)
 
--- a/src/nwm_routing/src/nwm_routing/__main__.py
+++ b/src/nwm_routing/src/nwm_routing/__main__.py
@@ -555,9 +555,9 @@ def _run_everything_v02(
     data_assimilation_folder = data_assimilation_parameters.get(
         "data_assimilation_timeslices_folder", None
     )
-    last_obs_file = data_assimilation_parameters.get("wrf_hydro_last_obs_file", None)
+    lastobs_file = data_assimilation_parameters.get("wrf_hydro_lastobs_file", None)
 
-    if data_assimilation_csv or data_assimilation_folder or last_obs_file:
+    if data_assimilation_csv or data_assimilation_folder or lastobs_file:
         if showtiming:
             start_time = time.time()
         if verbose:
--- a/src/nwm_routing/src/nwm_routing/__main__.py
+++ b/src/nwm_routing/src/nwm_routing/__main__.py
@@ -521,6 +521,15 @@ def _run_everything_v02(
         print("setting channel initial states ...")
 
     q0 = nnu.build_channel_initial_state(restart_parameters, param_df.index)
+    # STEP 4a: Set Channel States and T0
+    if restart_parameters.get("wrf_hydro_channel_restart_file",None):
+        channel_initial_states_file = restart_parameters["wrf_hydro_channel_restart_file"]
+        t0_str = nhd_io.get_param_str(channel_initial_states_file, "Restart_Time")
+    else:
+        t0_str = "2015-08-16_00:00:00"
+
+    t0 = datetime.strptime(t0_str, "%Y-%m-%d_%H:%M:%S")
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/nwm_routing/src/nwm_routing/preprocess.py [hunk 1]
```diff
@@ -206,31 +198,38 @@ def nwm_initial_warmstate_preprocess(
...
     if verbose:
         print("setting channel initial states ...")
 
     q0 = nnu.build_channel_initial_state(restart_parameters, segment_index)
 
+    if restart_parameters.get("wrf_hydro_channel_restart_file",None):
+        channel_initial_states_file = restart_parameters["wrf_hydro_channel_restart_file"]
+        t0_str = nhd_io.get_param_str(channel_initial_states_file, "Restart_Time")
+    else:
+        t0_str = "2015-08-16_00:00:00"
```
jdmattern-noaa, line 223: @jameshalgren I will use this t0. Do you have a method to get t0 for a cold start?

jameshalgren: In short, no. See #427 for additional detail on updating the t0 read.

awlostowski-noaa:
In the testing, V2 and V3 appear to produce different results at 166737579 with DA OFF.

src/nwm_routing/src/nwm_routing/__main__.py
```diff
@@ -941,6 +953,23 @@ def get_waterbody_water_elevation(waterbodies_df, q0):
...
+            )
+            for rr in run_results
+            if not rr[3][0].size == 0
+        ],
+        copy=False,
+    )
+    df["time_since_lastobs"] = df["time_since_lastobs"] - time_increment
+    return df
+
 def main_v03(argv):
```
awlostowski-noaa, line 973: This function could use some commenting to explain the sequence of processes... I'll suggest (based on my understanding) in comments below...

src/nwm_routing/src/nwm_routing/__main__.py
```diff
@@ -941,6 +953,23 @@ def get_waterbody_water_elevation(waterbodies_df, q0):
...
     args = _handle_args_v03(argv)
```
awlostowski-noaa, line 974:
```
# read input arguments from yaml\
```

awlostowski-noaa (commented):
main_v03 could benefit from more comments to help users understand the sequence of events.

src/nwm_routing/src/nwm_routing/preprocess.py
```diff
@@ -206,36 +199,47 @@ def nwm_initial_warmstate_preprocess(
                 len(waterbodies_initial_states_df)
             )
 
+        waterbodies_df = pd.merge(
```
jdmattern-noaa, line 202: This checks out to me as it is just the same thing you have in main at line 508.

src/nwm_routing/src/nwm_routing/preprocess.py
jdmattern-noaa: I actually don't need this t0.

jdmattern-noaa (approved):
This looks good to me on the reservoir side. The vast majority looks to be DA related that I am unfamiliar with here.

jameshalgren:
> In the testing, V2 and V3 appear to produce different results at 166737579 with DA OFF.

A closer review of testing confirmed a difference resulting from the interaction of interpolation of data assimilation values through a large gap across the break between looping segments. The linear interpolation respects the long-term trend across the gap, even though it will only fill the values as much as permitted by the `max_fill` parameter. With the looping tests above, only exactly the number of files needed for the given loop segment were passed to the loop. 

With commit https://github.com/NOAA-OWP/t-route/pull/429/commits/9a0b7a885b35b36db1ec1e31c43cf692f6f338f7, the input timeslice lists may now be padded on both sides an arbitrary amount to ensure that interpolation for the looped chunks is the same as it would be if the simulation were performed without breaks. 

Please compare these results to the DA_ON results from Below Falls Lake, segment 166737579, in the PR description above. There is still a difference in the values less than the floating point precision of the variables, but otherwise the values are the same. 
image
image

@awlostowski-noaa

awlostowski-noaa (dismissed):

src/nwm_routing/src/nwm_routing/__main__.py
jameshalgren: added a comment to this effect.

src/nwm_routing/src/nwm_routing/__main__.py
jameshalgren: Good suggestion. Expanded the doc-string with these details.

awlostowski-noaa (approved):
perfect - thanks for this addition

PR #430, opened 2021-09-14
v1: compute initial Y, C, D from iniq for CNT

kumdonoaa:
This PR adds the capability to initialize depth, celerity and diffusivity as a function of initial flow conditions in the diffusive CNT model. Prior to this PR, initial flow conditions were being passed from t-route framework and utilized to populate initial flow arrays in the Fortran kernel. However, depth, celerity and diffusivity were being arbitrarily initialized at the first timestep. These changes physically align all initial channel states.

Issue #424, opened 2021-09-13
Diffusive wave initial depth, celerity, and diffusivity conditions in CNT code


awlostowski-noaa:
Generate initial depth, celerity and diffusivity conditions from ini_q array. We have already done this for the Tulane code, need to port that capability to here at the kernel-level.

PR #432, opened 2021-09-15
Data Assimilation (DA) for the diffusive wave model(s)

awlostowski-noaa:
# Executive Summary
This PR adds the capability to do data assimilation (DA) with the diffusive wave model. At gage locations, simulated flows are either replaced with interpolated USGS observations, or nudged by some amount (in the event that observations are missing). Replaced/nudged flows at gage locations are then routed downstream. The DA scheme does not propagate any gage information upstream. Instead gage information (flow) is only passed in the downstream direction. The approach used for observation interpolation, DA substitution, and DA decay is exactly the same as is used with Muskingum-Cunge. 

# Technical Details
- DA with the diffusive wave model is facilitated by subnetwork creation. The spatial unit of analysis for the diffusive wave Fortran kernel is a stand-alone stream network, as opposed to a single segment in the case of Muskingum-Cunge. Therefore, it is necessary to break networks into subnetworks that reside between gages and reservoirs. Subnetwork tailwaters are either gages, reservoir inflows, or network tailwaters. This allows downstream flow conditions from one subnetwork to be passed as upstream boundary conditions to the subsequent subnetwork (either a stream network or reservoir). Subnetwork creation between gages and reservoirs is done by revisions in [`nhd_network.build_subnetworks_btw_reservoirs`](https://github.com/awlostowski-noaa/t-route/blob/5c9fed318caeef2588131b8c2b1da49ee8af3422/src/python_framework_v02/troute/nhd_network.py#L610-L765). When handed a list of gages in the network (from `usgs_df`) the function builds subnetworks between reservoirs and gages. Otherwise, the function only build subnetworks between reservoirs. 
- The DA substitution and decay processes are very similar to that used with Muskingum-Cunge, in that they both use the [`simple_da.simple_da_with_decay`](https://github.com/awlostowski-noaa/t-route/blob/5c9fed318caeef2588131b8c2b1da49ee8af3422/src/python_routing_v02/troute/routing/fast_reach/simple_da.pyx#L81-L117) function called from a cython compute kernel. However, the DA scheme differs from Muskingum-Cunge in that the diffusive DA occurs [AFTER](https://github.com/awlostowski-noaa/t-route/blob/5c9fed318caeef2588131b8c2b1da49ee8af3422/src/python_routing_v02/troute/routing/fast_reach/diffusive_cnt.pyx#L255-L315) all segments have been simulated for all timesteps. 
- [Changes to `diffusive_utils.py`](https://github.com/awlostowski-noaa/t-route/blob/5c9fed318caeef2588131b8c2b1da49ee8af3422/src/python_routing_v02/troute/routing/diffusive_utils.py#L500-L523) allow multiple off-network upstream inflows to a single subnetwork headwater as an upstream boundary condition. Gages frequently exist on junction inflow segments. So, the segment immediately below that junction (subnetwork headwater), needs to receive boundary condition flows from the gage segment and all other segments influent to the junction.  

# Demonstration
The figure below demonstrates diffusive wave data assimilation in action. The mainstem of the Eno River, between Hillsborough and Durham, North Carolina has two gages. Gage segments are colored red. Above the upstream most gage at Hillsborough, the open loop simulation is shown - this solution is replaced at the gage segment near Hillsborough. Here, we can see that the assimilation routine is properly replacing routed flows with interpolated observations. Between the two gages, the assimilated observations are routed downstream. At the downstream gage, the simulations are again replaced with interpolated observations.
diffusive_da_demo_eno

# Testing
1. Ensure that simulation results are stable throughout the domain - no negative or NaNs in the result arrays
- Results are stable. With a dt parameter of 300 seconds, the diffusive_cnt code exhibits numerical oscillations at headwater locations. Though with a dt = 600 seconds, this behavior is minimized. Despite some numerical oscillations, the simulation does not blow up at any point. Numerical oscillations near headwaters are ultimately less concerning because the diffusive wave model is not intended for use in headwater areas. 
2. Ensure that DA is happening correctly at gage locations - should see same exact substitution result and similar nudge pattern as with Muskingum-Cunge
- passed
3. Results for all segments and reservoirs above gages should be identical to no DA simulation result
- passed

## TODO
- [x] copy capability to Tulane diffusive version, not just CNT
- [x] proof of concept figures/charts
- [x] comprehensive explaination
- [x] test that workflow still works when no DA parameters are specified in YAML (i.e. dw with reservoirs, no da)
- [x] clean cython code
- [x] use upstream boundary data at native time resolution - why sub-sample to timestep of qlaterals?
- [x] what is going on when nts is set to limit of available CHRTOUT files?

### DIFF
--- a/src/python_routing_v02/troute/routing/compute.py
+++ b/src/python_routing_v02/troute/routing/compute.py
@@ -636,8 +636,9 @@ def compute_nhd_routing_v02(
             print("PARALLEL TIME %s seconds." % (time.time() - start_para_time))
 
     elif parallel_compute_method == "by-subnetwork-diffusive":
+        gages = set(usgs_df.index)
         reaches_ordered_bysubntw, subnetworks, subnetworks_only_ordered_jit = nhd_network.build_subnetworks_btw_reservoirs(
-            connections, rconn, wbody_conn, independent_networks, sources=None
+            connections, rconn, wbody_conn, gages, independent_networks, sources=None
         )
 
         if 1 == 1:
--- a/src/python_routing_v02/troute/routing/compute.py
+++ b/src/python_routing_v02/troute/routing/compute.py
@@ -636,11 +636,11 @@ def compute_nhd_routing_v02(
             print("PARALLEL TIME %s seconds." % (time.time() - start_para_time))
 
     elif parallel_compute_method == "by-subnetwork-diffusive":
-        gages = set(usgs_df.index)
+        gages = set(usgs_df.index).intersection(set(param_df.index))
         reaches_ordered_bysubntw, subnetworks, subnetworks_only_ordered_jit = nhd_network.build_subnetworks_btw_reservoirs(
             connections, rconn, wbody_conn, gages, independent_networks, sources=None
         )
-
+        
         if 1 == 1:
             print("JIT Preprocessing time %s seconds." % (time.time() - start_time))
             print("starting Parallel JIT calculation")
--- a/src/fortran_routing/diffusive_v02frwk/diffusive.f90
+++ b/src/fortran_routing/diffusive_v02frwk/diffusive.f90
@@ -675,16 +675,29 @@ contains
             !        print*, t,i,j,newQ(i,j),newY(i,j)-z(i,j)
             !	 enddo
 	    !end do
+            ! write results, timestep 2 and beyond
             if ( (mod( (t-t0*60.)*60.  ,saveInterval) .le. TOLERANCE) .or. ( t .eq. tfin *60. ) ) then
                 do j = 1, nlinks
                     ncomp= frnw_g(j,1)
                     do i=1, ncomp
-                        q_ev_g(ts_ev, i, j)= newQ(i,j)
-                        elv_ev_g(ts_ev, i, j)= newY(i,j)
+                        q_ev_g(ts_ev+1, i, j)= newQ(i,j)
+                        elv_ev_g(ts_ev+1, i, j)= newY(i,j)
                     enddo
                 enddo
                 ts_ev=ts_ev+1
             end if
+            
+            ! write initial conditions - timestep 1
+            if ( ( t .eq. t0 + dtini/60 ) ) then
+                do j = 1, nlinks
+                    ncomp= frnw_g(j,1)
+                    do i=1, ncomp
+                        q_ev_g(1, i, j)= oldQ(i,j)
+                        elv_ev_g(1, i, j)= oldY(i,j)
+                    enddo
+                enddo
+            end if
+            
               ! update of Y, Q and Area vectors
             oldY   = newY
             newY=-999
--- a/src/fortran_routing/diffusive_v02frwk/diffusive_cnt/diffusive_cnt.f90
+++ b/src/fortran_routing/diffusive_v02frwk/diffusive_cnt/diffusive_cnt.f90
@@ -43,7 +43,7 @@ contains
 
 		implicit none
 
-		integer, intent(in) :: mxncomp_g, nrch_g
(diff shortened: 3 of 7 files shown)


awlostowski-noaa:
ping @kumdonoaa for your awareness

awlostowski-noaa:
@kumdonoaa 
Following the discussion in #435, I revised the handling of upstream boundary conditions in `diffusive_utls`. Specifically, at t0 (initial condition), the sum of initial upstream flows is specified as a boundary condition in `ubcd_g`. Now, `ubcd_g` is an n x m array, where n is the same as `ntss_ev_g`, the number of written-out simulation timesteps, which includes the initial condition, and m is the number of headwater nodes in the model domain. The first row of the `ubcd_g` matrix contains the initial conditions of the all nodes upstream of the boundary node (offnetwork_upstream). Ultimately, this change improves the time-alignment of the diffusive wave model and it's boundary conditions - which may come from Muskingum-Cunge, a reservoir module, or a data-assimilated diffusive wave simulation, depending on model configuration. 

## Specific change:
Most of the changes associated with this comment are contained in [this commit](https://github.com/NOAA-OWP/t-route/pull/432/commits/463940b3062722d1d0595a8deee85a3cb9ed43b7). 

- Expand the size of the `upstream_flows_array` to n, m - where n is the number of segments downstream of an offnetwork-upstream inflow and and m is `nsteps+1`
- Get reservoir initial conditions from reservoir parameter array, which is newly passed down to the `diffusive_utils` function
- write cumulative upstream initial conditions to `upstream_flows_array`

awlostowski-noaa:
Unlike the Muskingum-Cunge and reservoir models, returned arrays of flow and depth contain data for the model initial states. (`nsteps+1`). In order to align the model output with that of other routing models in the framework, the first three columns of the `flowveldepth` array are dropped upon return to `compute.py`.  This drops the initial conditions and returns `nsteps` timesteps.

https://github.com/NOAA-OWP/t-route/pull/432/commits/dbdeb3436cc1dca416236d41490b1ae1817cf087#diff-7fadfa2dfa6aaa38429220877b8414d3862b5384944e1d11e1ef6432947565dfR325

awlostowski-noaa:
Investigation of the CNT code revealed that it was not returning the final computation timestep in the output arrays, `q_ev_g` and `elv_ev_g`. To fix this, a change is made to the loop indexing for writing model outputs. Specifically, the do-loop iterates from 1 to `repeatInterval+1`. This causes the final calculation timestep of the `repeatInterval` to be written over in `q_ev_g` and `elv_ev_g` for all but the final set of `repeatIntervals` (so somewhat inefficient), but ultimately results in the final computation timestep being written to output arrays. 

https://github.com/NOAA-OWP/t-route/pull/432/commits/46d8209ca3ba20c0c53c59953b2320c56da5971e

awlostowski-noaa:
Initialization of depth, celerity and diffusivity is added to `diffusive_cnt.f90`. Manually incorporated changes made by @kumdonoaa in #430. Will close #430. Close #424

src/python_routing_v02/troute/routing/compute.py
```diff
@@ -636,10 +636,11 @@ def compute_nhd_routing_v02(
             print("PARALLEL TIME %s seconds." % (time.time() - start_para_time))
 
     elif parallel_compute_method == "by-subnetwork-diffusive":
+        gages = set(usgs_df.index).intersection(set(param_df.index))
```
awlostowski-noaa, line 639: Is there another source of this set already created in compute.py?

awlostowski-noaa (commented):
@kumdonoaa 
I edited the printing logic so the initial condition is also included in the print.

jameshalgren (approved):
Added #442 as a suggestion to de-mystify the variable names.

PR #439, opened 2021-09-20
Last obs write out

jhrehanoaa:
I was testing this on Florence_test2_loop.yaml on Cheyenne. It appears to be working, but does not maintain parity over time. I believe this is caused by the calculation being used for how far back since the last obs was observed that comes from the "results". you can specify an output folder to send the lastobs csvs to or not and it will write to whatever the current directory is. Lastobs_df is recreated each following iteration and is updated with the correct gages and q0 values. Lastobs_df can then be fed back into the next run in the same format as the initial run. 

The current changes take place in main.py and the Florence_test2_loop.yaml files. It looks like some formatting changes caused some other files to be posted in this WIP. Added a new yaml called Florence_test_lastobs_writeout.yaml on Cheyenne for simplicity.


jhrehanoaa:
Last obs netcdf write out #457 will replace this PR.

Issue #444, opened 2021-09-22
DA coldstart with NaN in initial condition column of usgs_df


awlostowski-noaa:
If the DA routine is cold started - meaning that we have TimeSlice but no LastObs data - AND there is an NaN value in the first column of the usgs_df, then that NaN value will be erroneously substituted. This error occurs in the Falls Lake domain downstream at the gage downstream of Falls Lake (link 166737579). 

In concept, if there is no observation at the initial condition, the DA scheme leverages the last observation to substitute a nudged simulation. However, in the absence of any LastObs data, the NaN is substituted. We need to change `simple_da` to accommodate this "cold start" case by doing no substitution or nudging, just leave the open-loop value in place. 

## suggested changes in `simple_da.pyx`
``` cython
    if isnan(target_val):
        if da_check_gage:
            printf("THIS IS A NAN\t")
    if ((timestep <= gage_maxtimestep) and not (isnan(target_val))):
        if da_check_gage:
            printf("replace\t")
        replacement_val = target_val
        nudge_val = target_val - model_val
        # add/update lastobs_timestep
        lastobs_time = (timestep) * routing_period
        lastobs_val = target_val
    # suggested insertion here vvvv
    if (isnan(target_val)) and lastobs_val is empty: # <---- not the correct syntax but the correct logic
        replacement_val = model_val
        ...
   # ^^^^^  
   else:
        if da_check_gage:
            printf("So we are fixing that...\t")
        if da_check_gage:
            printf("decay; target: %g\t", target_val)
        da_decay_minutes = ((timestep) * routing_period - lastobs_time) / 60 # seconds to minutes
        da_weighted_shift = obs_persist_shift(lastobs_val, model_val, da_decay_minutes, decay_coeff)
        nudge_val = da_weighted_shift
```

## Documenting Issue Behavior
Parity check at 166737579 with TimeSlice files but no LastObs:
```
Comparing flows at 166737579
                     flow, wrf (cms)  flow, t-route (cms)  diff  rel_diff  absdiff  rel_absdiff
2018-09-17 01:00:00         7.038952                  NaN   NaN       NaN      NaN          NaN
2018-09-17 02:00:00         7.077681                  NaN   NaN       NaN      NaN          NaN
       flow, wrf (cms)  flow, t-route (cms)  diff  rel_diff  absdiff  rel_absdiff
count         2.000000                  0.0   0.0       0.0      0.0          0.0
[...]

PR #445, opened 2021-09-22
DA cold start fix

awlostowski-noaa:
This pull request proposes a fix to `simple_da` for cold-start DA situations. Specifically, when there is no LastObs file provided and there is a NAN in the first column of the usgs_df, the DA scheme will simply "substitute" the modeled value (i.e. no substitution and a nudge of zero). Addressing #444

### DIFF
--- a/src/python_routing_v02/troute/routing/fast_reach/simple_da.pyx
+++ b/src/python_routing_v02/troute/routing/fast_reach/simple_da.pyx
@@ -1,6 +1,5 @@
-from libc.math cimport exp, isnan
+from libc.math cimport exp, isnan, NAN
 from libc.stdio cimport printf
-
 
 cpdef float simple_da_with_decay_py(
     const float last_valid_obs,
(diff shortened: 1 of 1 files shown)


src/python_routing_v02/troute/routing/fast_reach/simple_da.pyx
```diff
@@ -51,6 +50,11 @@ cdef (float, float, float, float) simple_da(
         # add/update lastobs_timestep
         lastobs_time = (timestep) * routing_period
         lastobs_val = target_val
+    if ((isnan(target_val)) and (isnan(lastobs_val))):
+        replacement_val = model_val
+        nudge_val = 0.0
+        lastobs_val = NAN
+        lastobs_time = NAN
```
awlostowski-noaa, line 57: @jameshalgren can you confirm that the order of these if statements followed by the else is correct?

jameshalgren: It was not correct... I will push a suggested correction.

jameshalgren (approved):
This is ok to merge -- but if we have time, this would be a stellar opportunity for a unit test. The unit test could feed a handful of different combinations of values or NaNs and make sure we see the right things coming out.

Issue #433, opened 2021-09-16
Data assimilation resampling is hard-wired at 5-minute interval


awlostowski-noaa:
# Issue concept
The interpolation and resampling us USGS observations in `nhd_io.get_usgs_from_time_slices_folder` is hard-wired at a 5-minute interval:
```python
    usgs_df_T = (usgs_df_T.resample('min').
                 interpolate(limit = max_fill_1min, limit_direction = 'both').
                 resample('5min').  # <------ hard-wired resample interval here!
                 asfreq().
                 loc[date_time_center_start:,:])
```
# Issue manifestation
If the dt specified in the simulation yaml is 300 secs (5 mins), then this is not a problem. However, if a user specifies a different dt, then the timing of observation substitution at gage locations is incorrect. For example, the figures below show simulation results with DA for a dt = 600 sec simulation. Above the gage (first figure), the open-loop t-route simulation (blue) is reasonable compared to WRF hydro (orange). However, at the gage location, the t-route simulation is completely incorrect next to WRF hydro and these erroneous results are then routed downstream. 

### Above gage - 10 minute timestep solution closely resembles WRF solution
![image](https://user-images.githubusercontent.com/66840445/133620708-ef3ba921-08e0-48ff-8e07-77ca19685d86.png)

### At the gage - 10 minute timestep solution is wrong
![image](https://user-images.githubusercontent.com/66840445/133620889-7aa6308f-77eb-41a2-b636-530a3b05d296.png)

# Proposed solution
use dt from yaml to control resampling interval 

@jameshalgren

jameshalgren:
Good point, @awlostowski-noaa. This TODO logs a point in the code where such a parameter might be read for later use in a more flexible interpolation scheme: 
https://github.com/NOAA-OWP/t-route/blob/4ec74cc70d4f1ca407d20a6764b3f89facb43a96/src/python_framework_v02/troute/nhd_network_utilities_v02.py#L690

jameshalgren:
@jhrehanoaa 
This needs to be connected to dt: 
https://github.com/NOAA-OWP/t-route/blob/4ec74cc70d4f1ca407d20a6764b3f89facb43a96/src/python_framework_v02/troute/nhd_io.py#L776

This needs to be an `interpolation_length`
https://github.com/NOAA-OWP/t-route/blob/4ec74cc70d4f1ca407d20a6764b3f89facb43a96/src/python_framework_v02/troute/nhd_io.py#L652

[...]

PR #447, opened 2021-09-22
allow different intervals for data assimilation

jhrehanoaa:
This PR is built off of the fixes found in jameshalgren/long-term-looping-all.

This PR should address the feature requests in:
**https://github.com/NOAA-OWP/t-route/issues/433**

marked at:

- https://github.com/NOAA-OWP/t-route/blob/4ec74cc70d4f1ca407d20a6764b3f89facb43a96/src/python_framework_v02/troute/nhd_io.py#L757

- https://github.com/NOAA-OWP/t-route/blob/4ec74cc70d4f1ca407d20a6764b3f89facb43a96/src/python_framework_v02/troute/nhd_io.py#L652

- https://github.com/NOAA-OWP/t-route/blob/4ec74cc70d4f1ca407d20a6764b3f89facb43a96/src/python_framework_v02/troute/nhd_io.py#L776

The frequency needed to be calculated using the run_parameters['dt']. I made a few changes to pass run_parameters through the necessary functions to get it there. Additionally, a slight modification to data_assimilation_parameters to add both data_assimilation_interpolation_limit: 14 and QC_threshold: 1. These can now be specified and passed from the YAML to the functions found in .io.

The files I made changes to are the Florence_fcast_da_5min, main, nhd_io, and nhd_network_utilities. Added new yaml named Florence_fcast_da_5min_frequency_test.yaml to the shared folder on Cheyenne for simplicity.

Some of the new t0 stuff James added is showing as new I am not sure why that is occurring as it has both been rebased and merged along with some lines that are the same. I think some of it may be from when I originally branched off of James's branch.

### DIFF
--- a/src/python_framework_v02/troute/nhd_io.py
+++ b/src/python_framework_v02/troute/nhd_io.py
@@ -666,10 +666,10 @@ def get_usgs_df_from_csv(usgs_csv, routelink_subset_file, index_col="link"):
 
 def get_usgs_from_time_slices_folder(
     routelink_subset_file,
-    run_parameters,
+    dt,
     usgs_files,
-    data_assimilation_parameters,
-    max_fill_1min = 14,
+    qc_trehsh,
+    max_fill_1min,
     t0 = None,
 ):
     """
@@ -682,9 +682,7 @@ def get_usgs_from_time_slices_folder(
           the interpolated values are truncated so that the first value returned
           corresponds to the first center date of the first provided file.
     """
-    max_fill_1min = data_assimilation_parameters.get("data_assimilation_interpolation_limit", 59)
-    qc_trehsh = data_assimilation_parameters.get("QC_threshold", 1)
-    frequency = str(int(run_parameters['dt']/60))+"min"
+    frequency = str(int(dt/60))+"min"
     with read_netcdfs(usgs_files, "time", preprocess_time_station_index,) as ds2:
 
         # dataframe containing discharge observations
--- a/src/python_framework_v02/troute/nhd_io.py
+++ b/src/python_framework_v02/troute/nhd_io.py
@@ -666,7 +666,9 @@ def get_usgs_df_from_csv(usgs_csv, routelink_subset_file, index_col="link"):
 
 def get_usgs_from_time_slices_folder(
     routelink_subset_file,
+    run_parameters,
     usgs_files,
+    data_assimilation_parameters,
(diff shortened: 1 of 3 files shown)


src/python_framework_v02/troute/nhd_io.py
```diff
@@ -680,6 +682,9 @@ def get_usgs_from_time_slices_folder(
           the interpolated values are truncated so that the first value returned
           corresponds to the first center date of the first provided file.
     """
+    max_fill_1min = data_assimilation_parameters.get("data_assimilation_interpolation_limit", 59)
+    qc_trehsh = data_assimilation_parameters.get("QC_threshold", 1)
```
jameshalgren, line 686: This is a spelling error we made in the original notes. Let's call this `qc_threshold` or `qc_thresh` for short, if needed.

src/python_framework_v02/troute/nhd_network_utilities_v02.py
```diff
@@ -715,8 +714,9 @@ def build_data_assimilation_folder(data_assimilation_parameters, t0=None):
 
     usgs_df = nhd_io.get_usgs_from_time_slices_folder(
         data_assimilation_parameters["wrf_hydro_da_channel_ID_crosswalk_file"],
+        run_parameters,
```
jameshalgren, line 717: Down at this level, I would recommend that we pass just the needed parameters as scalar values rather than as a dict. 
Same as comment in nhd_io.

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -666,7 +666,9 @@ def get_usgs_df_from_csv(usgs_csv, routelink_subset_file, index_col="link"):
 
 def get_usgs_from_time_slices_folder(
     routelink_subset_file,
+    run_parameters,
```
jameshalgren, line 669: At this level, let's pass explicitly what we need as scalar values with no dict parameters. That makes this function more useful from other contexts and makes it super clear what we need out of those dictionaries. Same comment as below in the calling function.

awlostowski-noaa:
Running a test in the Florence domain with `dt = 600` and `qts_subdivisions = 6`, expect to see parity with WRF Hydro at gage locations with complete records on the hour. Expectations are met!

awlostowski-noaa (approved):
All looks good to me. Behaves as I expected, too!

PR #454, opened 2021-09-28
bugfix: modify v03 exec pathway for da updates

jameshalgren:
Changes in nhd_network_utilities_v02.py #447 are applied here to the v03 method of execution (i.e., "looping" execution).


awlostowski-noaa (approved):
All good. Please merge this before reviewing #456.

PR #456, opened 2021-09-30
Update diffusive DA to enable V3 exec pathway

awlostowski-noaa:
In order to enable simulations through the looped V3 execution pathway, we need to pass LastObs data from cython to `compute.py`. Our first implementation of DA in `diffusive.pyx` and `diffusive_cnt.pyx` did not gather and return LastObs data, which caused the V3 execution pathway to fail.

### DIFF
--- a/src/python_routing_v02/troute/routing/fast_reach/diffusive.pyx
+++ b/src/python_routing_v02/troute/routing/fast_reach/diffusive.pyx
@@ -273,7 +273,7 @@ cpdef object compute_diffusive_tst(
     cdef int gages_size = usgs_positions.shape[0]
     cdef int gage_maxtimestep = usgs_values.shape[1]
     cdef int gage_i, usgs_position_i, timestep
-    cdef float lastosbs_value, lastobs_time
+    cdef float lastobs_value, lastobs_time
     cdef float a, da_decay_minutes, da_weighted_shift, replacement_val  # , original_val, lastobs_val,
     cdef float [:] lastobs_values, lastobs_times
     cdef (float, float, float, float) da_buf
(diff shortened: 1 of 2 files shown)


awlostowski-noaa:
To test the V3 execution pathway with the diffusive wave model, run these two .yaml files (with this branch checked out). You will absolutely need to recompile before running, as all of the changes are in cython code. 

- /glade/u/home/adamw/projects/t-route/test/jobs/V03/Flat_River_test_CNmod_V3_loop.yaml
- /glade/u/home/adamw/projects/t-route/test/jobs/V03/Tar_River_test_CNT_V3.yaml

awlostowski-noaa:
## known issue
Due to how time is handled in the CNT model, the looped solution does not match a non-looped solution for the timesteps immediately following the start of a loop iteration. In a nutshell, all flow computations in a `repeatInterval` are calculated assuming the same channel depth, diffusivity, and celerity state. We are in the process of addressing this issue to changing the model to calculate depth, diffusivity, and celerity at each timestep that flow is calculated. While this outstanding issue affects model accuracy - it does not affect stability.

jameshalgren:
Merged #454 -- rebase is clean on top of that. 
Test files both execute -- Flat river is completely clean; Tar river runs with a few depth iteration warnings.

src/python_routing_v02/troute/routing/fast_reach/diffusive.pyx
```diff
@@ -273,21 +273,25 @@ cpdef object compute_diffusive_tst(
     cdef int gages_size = usgs_positions.shape[0]
     cdef int gage_maxtimestep = usgs_values.shape[1]
     cdef int gage_i, usgs_position_i, timestep
-    cdef float lastosbs_value, lastobs_time
+    cdef float[:] lastosbs_values, lastobs_times
```
jameshalgren, line 276: This looks like it may be a spelling error...

awlostowski-noaa: Sure is. Do you mind fixing and committing?

jameshalgren (approved):
passes all new tests.

PR #457, opened 2021-09-30
Last obs netcdf write out

jhrehanoaa:
I was testing this on Florence_test2_loop.yaml on Cheyenne. This picks up where the csv write out left off. I created a new branch as there was some rebasing and merging issues with the previous PR branch. Essentially, this makes the lastobs_df each run with these columns requesting (time,flow,nudge,etc.) then reads and writes out to nc. 

The current changes take place in main.py and the Florence_test2_loop.yaml files. Added a new yaml called Florence_test_lastobs_writeout.yaml on Cheyenne for simplicity.


jhrehanoaa:
There appears to be a large section of async code removed despite us just recently merging with master. @jameshalgren I can go back and copy back in these differences.

jhrehanoaa:
Broken commit fix. Reopening.

PR #458, opened 2021-09-30
Write last obs to netcdf

jhrehanoaa:
I was testing this on Florence_test2_loop.yaml on Cheyenne. This picks up where the csv write out left off. I created a new branch as there was some rebasing and merging issues with the previous PR branch. Essentially, this makes the lastobs_df each run with these columns requesting (time,flow,nudge,etc.) then reads and writes out to nc. 

The current changes take place in main.py and the Florence_test2_loop.yaml files. Added a new yaml called Florence_test_lastobs_writeout.yaml on Cheyenne for simplicity.


jameshalgren:
Got this rebased with only relevant commits. @jhrehanoaa, please `git fetch && git checkout upstream/lastobssaved` and confirm prior testing.

src/nwm_routing/src/nwm_routing/__main__.py
```diff
@@ -1177,6 +1195,22 @@ def main_v03(argv):
...
+                else:
+                    lastobs_string = "lastobs_df_"+str(run_set_iterator)+".nc"
+                if not 'modelTimeAtOutput' in lastobs_df.columns:
+                    lastobs_df.insert(loc=0, column='modelTimeAtOutput', value=(time.time() - main_start_time))
+                if not 'Nudge' in lastobs_df.columns:
+                    lastobs_df.insert(loc=0, column='Nudge', value=np.NaN)
+                lastobs_df['Nudge'] = np.NaN
+                lastobs_df['modelTimeAtOutput'] = (time.time() - main_start_time)
+                lastobs_df.to_xarray().to_netcdf(lastobs_string)
+                # lastobs_df = pd.read_csv(lastobs_string,index_col='link')
```
jameshalgren, line 1213: It seems like we are doing the same thing twice with the code block up top and this one here. Could they be consolidated?

src/nwm_routing/src/nwm_routing/__main__.py
```diff
@@ -1125,6 +1128,21 @@ def main_v03(argv):
...
+                lastobs_string = "lastobs_df_"+str(run_set_iterator)+".nc"
+            if not 'modelTimeAtOutput' in lastobs_df.columns:
+                lastobs_df.insert(loc=0, column='modelTimeAtOutput', value=(time.time() - main_start_time))
+            if not 'Nudge' in lastobs_df.columns:
+                lastobs_df.insert(loc=0, column='Nudge', value=np.NaN)
+            lastobs_df['Nudge'] = np.NaN
+            lastobs_df['modelTimeAtOutput'] = (time.time() - main_start_time)
+            lastobs_df.to_xarray().to_netcdf(lastobs_string)
+            lastobs_df = xr.open_dataset(lastobs_string).to_dataframe()
+            lastobs_df['gages'] = lastobs_df['gages'].astype(str).str.decode('utf-8')
```
jameshalgren, line 1145: This block, as an output function, is out of place up here in this part of the execution loop. I would suggest we make a function out of the block, then put it inside of the general output call here: https://github.com/NOAA-OWP/t-route/blob/b6cace768be819889d48e4bc99924208da1b3180/src/nwm_routing/src/nwm_routing/__main__.py#L1187

jameshalgren: which... would remove the need for the xarray library in this main function.

jameshalgren (changes requested):
We need to make this functionality available to all three execution modes. Right now it is attached to `main_v03` (i.e., "looping") execution. It needs also to be attached to the `main_v03_async` probably here: https://github.com/NOAA-OWP/t-route/blob/b6cace768be819889d48e4bc99924208da1b3180/src/nwm_routing/src/nwm_routing/__main__.py#L1481 and to the `main_v02` execution path, probably here: https://github.com/NOAA-OWP/t-route/blob/b6cace768be819889d48e4bc99924208da1b3180/src/nwm_routing/src/nwm_routing/__main__.py#L734

jhrehanoaa:
I made the requested changes. A good amount of information has to be passed to those output functions. I also fetched both the origin and my own, but it looks like it went ahead and added the commits again.

jhrehanoaa:
replaced by #460 to remove extra commits

awlostowski-noaa:
superseded by #460

PR #460, opened 2021-10-01
Lastobs nc write out with naming convention

jhrehanoaa:
LastObs data are written out as netcdf at the end of each loop in a V4 or V3 execution, or at the end of the simulation in a V2 execution. LastObs writing only occurs if the `lastobs_output_folder` variable is provided under `data_assimilation_parameters` in the .yaml file.

PR #467, opened 2021-10-10
Lastobs write out final

jameshalgren:
Please feel free to compare to #460 ...

This is how I would suggest you make these changes.

### DIFF
--- a/src/python_framework_v02/troute/nhd_io.py
+++ b/src/python_framework_v02/troute/nhd_io.py
@@ -1,5 +1,4 @@
 import zipfile
-import warnings
 import xarray as xr
 import pandas as pd
 import geopandas as gpd
@@ -11,7 +10,6 @@ import dask.array as da
 import sys
 import math
 from datetime import *
-import time
 
 from troute.nhd_network import reverse_dict
 
@@ -661,10 +659,9 @@ def get_usgs_df_from_csv(usgs_csv, routelink_subset_file, index_col="link"):
         df = ds.to_dataframe()
 
     usgs_df = df.join(df2)
-    link_gage_df = usgs_df[['gages']]
     usgs_df = usgs_df.drop(["gages", "ascendingIndex", "to"], axis=1)
 
-    return usgs_df, link_gage_df
+    return usgs_df
 
 
 def get_usgs_from_time_slices_folder(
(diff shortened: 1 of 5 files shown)


src/python_framework_v02/troute/nhd_network_utilities_v02.py
```diff
@@ -714,10 +715,10 @@ def build_data_assimilation_folder(data_assimilation_parameters, run_parameters)
 
     usgs_df = nhd_io.get_usgs_from_time_slices_folder(
         data_assimilation_parameters["wrf_hydro_da_channel_ID_crosswalk_file"],
-        run_parameters["dt"],
         usgs_files,
         data_assimilation_parameters.get("qc_threshold", 1),
         data_assimilation_parameters.get("data_assimilation_interpolation_limit", 59),
+        run_parameters["dt"],
```
jameshalgren, line 721: None of the the changes in this file are germane to the changes here and perhaps should be omitted.

jameshalgren:
This branch still creates an infinite loop in the v03 and async methods of execution.

src/python_framework_v02/troute/nhd_io.py
```diff
@@ -1085,3 +1085,38 @@ def build_coastal_ncdf_dataframe(coastal_ncdf):
...
+        {
+            "stationId": (["stationIdInd"], lastobs_df["gages"].to_numpy(dtype = '|S15')),
+            "time": (["stationIdInd"], np.asarray(lastobs_timestamp_str,dtype = '|S19')),
+            "discharge": (["stationIdInd"], lastobs_df["lastobs_discharge"].to_numpy()),
+        }
+    )
+    ds.attrs["modelTimeAtOutput"] = "example attribute"
+
+    # write-out LastObs file as netcdf
+    ds.to_netcdf(lastobs_output_folder + "nudgingLastObs." + modelTimeAtOutput_str + ".nc")
```
jameshalgren, line 1122: Need to use pathlib here to join this path -- ended up with the humorous: 
`outputnudgingLastObs.2018-09-01_01:00:00.nc`

jameshalgren:
> getting gage ID data from the output of `nnu.build_network_connections` is definitely more efficient than the previously proposed approach that added output to the processes of building `usgs_df`. However, if `gages` is not listed as a dictionary value in the .yaml file, than the lastobs write process will fail. This is potentially a place where a warning message can help the customer understand what is missing and how to get around this error.
> 
> Also, I am unable to reproduce the infinite loop in my simple yaml test file: /glade/u/home/adamw/test/jobs/lastobs_write_test/Tar_River_test_MC_V3.yaml

Yes -- the gages parameter would have to be in the list and a warning would make sense. Without the gages parameter, the DA is going to break anywhere a gage is not at the bottom of the reach. It will still assimilate the values but the propagation of the values will be broken. (So that is another reason to include the warning.)

PR #470, opened 2021-10-12
all .py files formatted with black

jhrehanoaa:
I just formatted everything with black that ends with .py. Ignore the name of the branch "removed_files" I was going to try removing those extra files that we will be taking care of in the new release but decided to just do a separate PR. 

I ran `find \. -name "*.py" -exec black {} \;` inside of /t-route for reference.


jhrehanoaa:
Entire directories to be removed. Command for formatting all files still above. 

troute/src/lookup_routing
troute/src/python_routing_v00
troute/src/python_routing_v01
troute/src/python_framework_v01

jameshalgren:
@shorvath-noaa 
This can probably be closed. If someone wants to run the black formatter on the files at some point that can be a new PR.

PR #498, opened 2021-11-09
Logging, configuration checks, and timing summary

awlostowski-noaa:
- the Python logging library is now used in lieu of print statements throughout the codebase. Logging level is set in the configuration file with parameter `log_level`
- A simple check on configuration file content is added and executed with logging level is set to DEBUG. At the moment, the input check simply checks for the existence of required files and directories specified in the YAML file. 
- The logging parameter list has been trimmed - `verbose` and `debuglevel` are no longer needed. 
- The `showtiming` parameter has been repurposed control the construction and printing of an execution time summary report. If `showtiming == True`, then a timing summary is reported along the V3 execution pathway (synchronous only). The timing statements previously reported are still available through the DEBUG logging level. Here is an example of the timing output for a 3 hour CONUS simulation with no DA:
```
************ TIMING SUMMARY ************
----------------------------------------
Network graph construction: 40.88 secs, 14.66 %
Initial condition handling: 1.01 secs, 0.36 %
Forcing array construction: 1.63 secs, 0.58 %
Routing computations: 219.81 secs, 78.83 %
Output writing: 15.52 secs, 5.56 %
```

### DIFF
--- a/doc/v3_doc.yaml
+++ b/doc/v3_doc.yaml
@@ -1,4 +1,3 @@
-# > python -m nwm_routing -f -V3 <filename>.yaml
 # ---------------
 # parameters controling logging messages throughout execution
 log_parameters:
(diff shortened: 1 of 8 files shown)


jhrehanoaa (approved):
Works and looks good.

PR #505, opened 2021-12-08
Write flow at gage locations to netCDF (CHANOBS) file

awlostowski-noaa:
This PR proposes new output routines for writing flow results at gage locations to "CHANOBS" files. These additions are motivated by NWM3.0 calibration needs. 

## Additions

- new function: `nhd_io.write_chanobs`, which writes flow results from segments co-located with stream gages to netCDF. If the  user-specified netCDF file does not yet exist, one is created. If the user-specified file already exists, flow results are simply appended to it. 
- New output arguments in configuration .yaml file specifying the directory and name of the netCDF file that results are to be written to. 
- New conditional call to `nhd_io.write_chanobs` in  `output.nwm_output_generator` that is activated if chanobs output parameters exist in configuration file. 

## Removals

- None

## Changes

- None

## Testing

1.  Ran a 5.9 year (`nts=622000`) calibration simulation in test basin 06846500 with 1-month loop chunks. Flow results at the catchment outlet (gage location) were written to `/glade/scratch/adamw/chanobs_dump/chanobs_test`. It took about 0.15 seconds to append flows to netCDF output after each loop iteration. Below is a few lines of code I used to open and plot the data:
``` python
import xarray as xr
import pandas as pd

dat = xr.open_dataset("/glade/scratch/adamw/chanobs_dump/chanobs_test").streamflow.to_dataframe()
dat = dat.reset_index().set_index('time')
dat.loc[:,'streamflow'].plot()
```
image

2. A parity check between current upstream master and this development branch confirmed that code additions proposed here do no affect flow, velocity, or depth results. 

## Todos

- Writing entire time-series to a single file is really convenient for analysis. Otherwise, we need to stitch together multiple CHRTOUT files or .csv files. Future work may consider allowing users to specify the segments from which flows are written to netCDF. 

## Checklist

- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

### DIFF
--- a/src/nwm_routing/src/nwm_routing/output.py
+++ b/src/nwm_routing/src/nwm_routing/output.py
@@ -2,7 +2,7 @@ import time
 import numpy as np
 import pandas as pd
 from pathlib import Path
-from datetime import datetime
+from datetime import datetime, timedelta
 import troute.nhd_io as nhd_io
 from build_tests import parity_check
 import logging
@@ -73,6 +73,7 @@ def nwm_output_generator(
     rsrto = output_parameters.get("hydro_rst_output", None)
     chrto = output_parameters.get("chrtout_output", None)
     test = output_parameters.get("test_output", None)
+    chano = output_parameters.get("chanobs_output", None)
 
     if csv_output:
         csv_output_folder = output_parameters["csv_output"].get(
@@ -81,7 +82,7 @@ def nwm_output_generator(
         csv_output_segments = csv_output.get("csv_output_segments", None)
 
     
-    if csv_output_folder or rsrto or chrto or test:
+    if csv_output_folder or rsrto or chrto or chano or test:
         
         start = time.time()
         qvd_columns = pd.MultiIndex.from_product(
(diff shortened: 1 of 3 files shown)


src/nwm_routing/src/nwm_routing/output.py
```diff
@@ -213,6 +214,24 @@ def nwm_output_generator(
...
         # usgs_df_filtered = usgs_df[usgs_df.index.isin(csv_output_segments)]
         # usgs_df_filtered.to_csv(output_path.joinpath("usgs_df.csv"))
+        
+    if chano:
+        
+        LOG.info("- writing t-route flow results at gage locations to CHANOBS file")
+        start = time.time()
+                
+        nhd_io.write_chanobs(
+            Path(chano['chanobs_output_directory'] + chano['chanobs_filepath']),
```
jhrehanoaa, line 224: Wasn't sure if this could be simplified to just one directory path?

src/nwm_routing/src/nwm_routing/output.py
```diff
@@ -213,6 +214,24 @@ def nwm_output_generator(
         LOG.debug("writing CSV file took %s seconds." % (time.time() - start))
         # usgs_df_filtered = usgs_df[usgs_df.index.isin(csv_output_segments)]
         # usgs_df_filtered.to_csv(output_path.joinpath("usgs_df.csv"))
+        
+    if chano:
+        
+        LOG.info("- writing t-route flow results at gage locations to CHANOBS file")
+        start = time.time()
+                
+        nhd_io.write_chanobs(
```
jhrehanoaa, line 223: One issue I saw with this method is that if the file already exists it will just be appended after the first time running. If I continue to rerun a yaml it will continuously grow the output file. Maybe it would be better to delete the file if it exists before writing to it once again?

jhrehanoaa (approved):
The results that are generated match the threshold in which we find acceptable. I mentioned two small issues one regarding directory chano simplification and another regarding the chano output file continuously growing after rerunning the yaml file. Beyond those issues, I do not see anything glaring. The results created from two different matching serial outputs sometimes do not match as well. This is already documented in another issue regarding serial vs parallel, but I thought it noteworthy that it also occurs on serial vs serial runs. Same floating point precision error as before though. I tested this over the Florence domain and used the dynamic qlat creation as well just to make sure it wasn't being caused by explicit file naming.

src/nwm_routing/src/nwm_routing/output.py
```diff
@@ -213,6 +214,24 @@ def nwm_output_generator(
...
+            Path(chano['chanobs_output_directory'] + chano['chanobs_filepath']), 
+            flowveldepth, 
+            link_gage_df, 
+            t0,
```
jhrehanoaa, line 227: The model output looks like this on my end just for reference. 

Dimensions:         (feature_id: 145, reference_time: 1, time: 936)
Coordinates:
  * time            (time) datetime64[ns] 2018-09-01T00:05:00 ... 2018-09-02
  * reference_time  (reference_time) datetime64[ns] 2018-09-01
  * feature_id      (feature_id) int32 3350287 3350831 ... 166737579 933020162
Data variables:
    streamflow      (time, feature_id) float32 ...
Attributes:
    model_initialization_time:  2018-09-01_00:00:00
    model_output_valid_time:    2018-09-01_00:05:00

src/nwm_routing/src/nwm_routing/output.py
awlostowski-noaa: I spoke with our NCAR colleagues and they confirmed that appending to existing files is the best way to go. Say a simulation composed of 10 loops fails half way through the fifth. We wouldn't want to throw out the results of the first 4 loops and start from the beginning. Instead, we can use restart files to launch a follow up simulation from the top of the fifth loop and continue appending to the same netCDF output file. Yes, this will duplicate some information in the netCDF, but the user can filter-our non unique dates on when they are analyzing information in the netCDF.

PR #509, opened 2022-01-04
Pythonize reservoir (hybrid) data assimilation ++

awlostowski-noaa:
# Overview
The primary objective of this pull request is re-writing of  the USGS/USACE hybrid reservoir data assimilation (DA) routines in Python from Fortran. The motivation for this PR is that the current Fortran-based hybrid reservoir DA code is incompatible with t-routes parallelization schemes. While the Fortran-based solution produces scientifically accurate results, performance is hampered by inefficient (duplicate) reading of TimeSlice files that contain gage observations to be assimilated. 

## What exactly is hybrid reservoir DA anyway?
- Hybrid reservoir DA is essentially persisting observed outflows below a reservoir for multiple timesteps, and using the observed outflows as part of the reservoir water balance equation (change in storage = inflow - outflow). In this way, reservoir storage states are updated with assimilated outflow observations.
- There are not many reservoirs CONUS-wide where this flavor of DA happens -only 74 USGS and 122 USACE reservoirs. 
- The utility of hybrid DA is two-fold. First, it can (potentially) make reservoir storage state estimates more accurate. Second, reservoir outflow rates are directly informed by data, and each assimilated observation is persisted forward in time anywhere between 1 hour and ~11 days.
## Solution status
Re-writing the DA routines to Python allowed for a massiver performance gain. For example, the compute time for a 28-hour simulation in the Neuse River basin of North Carolina with hybrid reservoir DA was taking ~120 seconds with the Fortran-based solution. The Python-based solution has an average compute time of ~5 seconds. 

However, there are some minor discrepancies in the DA solutions. Specifically, the assimilation timing in the Python solution appears to be off by one timestep, compared to the Fortran solution. Here is an example of that at a USGS reservoir called Lake Michie in North Carolina. Notice that each scheme is assimilating and persisting the same flow values, however, the timing of the assimilation change is one timestep earlier in the Python solution.
image
Also, there are some differences in the between DA solutions in the case where reservoir storage states are exceeded. Again, both solutions have the same general behavior, but there appears to be a phase shift driving differences. Here is an example of that in the outflows from Buckhorn Reservoir, a USGS reservoir in North Carolina. Both solutions are flipping between the open-loop solution and the assimilated flow value. The open loop solution taken whenever the reservoir storage state is exceeded, which allows the storage to reduce more rapidly. As soon as the storage capacity is below the maximum threshold, the assimilated outflow is taken until the storage capacity is exceeded, at which point the outflow goes back to the level pool solution. 
image
Overall, the new Python implementation of reservoir DA is far from a perfect representation of the Fortran solution. However, it is much faster and captures essentials of the DA scheme. Getting this merged in will allow NCAR to proceed with testing at long-term CONUS scales, and we can update the Python DA code for accuracy and performance as needed. 

## Additions

- Python hybrid reservoir DA functions - [`reservoir_hybrid_da.py`](https://github.com/awlostowski-noaa/t-route/blob/res_da_debug/src/python_routing_v02/troute/routing/fast_reach/reservoir_hybrid_da.py)
- New and modified framework functions to read and organize TimeSlice data for hybrid data assimilation.
- 

## Changes

- Use `t0` to inform the `model_start_time` variable, which was previously being calculated from the column headers in the `qlats` dataframe. There is no guarantee that the header columns in qlat will be datetimes (in fact it is unlikely that they will be). If the headers are not datetimes then the `model_start_time` variable is set to something near current system time and the reservoir DA sequence does no operate correctly.  
- The structure of the configuration file has changed. We moved reservoir DA parameters under the `data_assimilation_parameters` section. There are two new sub-sections of the `data_assimilation_parameters` section - `streamflow_da` and `reservoir_da`, which contain parameters specific to streamflow and reservoir, DA respectively. 
- The writing and reading of `lastobs` files has been changed. The reading function, [`build_lastobs_df`](https://github.com/awlostowski-noaa/t-route/blob/e8e18177b03375db2e0e09b38a861dd13252469f/src/python_framework_v02/troute/nhd_io.py#L802-L904) has been modified to be more efficient and also support the reading of t-route generated lastobs files. On the write-side, the structure of the netCDF files created by [`lastobs_df_output`](https://github.com/awlostowski-noaa/t-route/blob/e8e18177b03375db2e0e09b38a861dd13252469f/src/python_framework_v02/troute/nhd_io.py#L1499-L1533) now contains all necessary variables and attributes to facilitate their use in subsequent t-route simulations. 

## Testing

1. Neuse River domain that contains two USGS hybrid reservoirs reservoirs: `/glade/u/home/adamw/projects/t-route/test/jobs/reservoir_debug/neuse.yaml`
2. CONUS domain full AnA simulation: `/glade/u/home/adamw/projects/t-route/test/jobs/reservoir_debug/conus_test/conus.yaml`

## Notes

- There is admittedly a fair bit of scope creep in this PR. Implementing the reservoir DA solution in Python required updates in many corners of the codebase. Along the way, when I took advantages of opportunities to make existing functions more efficient, add docstrings, and improve parameter defaults. 

## Todos

-

## Checklist

- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)


awlostowski-noaa:
@jhrehanoaa I found this issue, which was leading to us seeing the subrouting `read_timeslice_file` never being called. Specifically, it caused this statement to evaluate as `FALSE` all the time because the `timeslice_file_name` was being constructed from an incorrect `old_date` variable value. 

``` Fortran
! Check if file exists
inquire(FILE = timeslice_file_name_full, EXIST = file_exists)
if (file_exists) then
  ! Call subroutine to read a particular timeslice file
  call this%read_timeslice_file(timeslice_file_name_full, timeslice_file_name)
end if
```
This is issue was created by changes in out I/O schemes from recent PRs that affected the column headers in the `qlats` DataFrame. Unfortunately, I didn't realize that the column headers were being relied upon to inform simulation time in the reservoir DA schemes. Now that we have this issue shored up, we should be able to go hunting for the TimeSlice I/O inefficiently that we think exists.

jhrehanoaa (dismissed):
Corrects model start time used in fortran subprocesses. Passes checks.

src/nwm_routing/src/nwm_routing/__main__.py
```diff
@@ -1198,6 +1204,7 @@ def main_v03(argv):
             compute_kernel,
             subnetwork_target_size,
             cpu_pool,
+            t0,
             dt,
             nts,
             qts_subdivisions,
```
jhrehanoaa, line 1210: After run_results is returned can we just copy over the values found in usgs or usace? There is only a fraction of the timestep locations, but this copies over the first 0:n timesteps. Possible approach.


#find the location of the reservoir ids in order to index the results values. 
  usgs_idx=np.where(np.isin(run_results[0][0], [reservoir_usgs_df.index]))[0]
  usace_idx=np.where(np.isin(run_results[0][0], [reservoir_usace_df.index]))[0]
  #loop through values and replace n timesteps with da values. 
  for location, value in enumerate(usgs_idx):
      run_results[0][1][value][:len(reservoir_usgs_df.values[location])] = reservoir_usgs_df.values[location]
  for location, value in enumerate(usace_idx):
      run_results[0][1][value][:len(reservoir_usace_df.values[location])] = reservoir_usace_df.values[location]

PR #515, opened 2022-02-07
Diffusive routing for both natural & RouteLink channel cross sections with stable depth computation

kumdonoaa:
[Short description explaining the high-level reason for the pull request]
## Overview

The main updates for this PR are
1. In diffusive.f90, as topobathy data represents channel cross section, in general, more accurately than RouteLink data, we enabled diffusive model to work with both topobathy and RouteLink data.    

2.  Topobathy data used here has manning's N values at all the vertices so that the values used for creating hydraulic lookup table should be composite or equivalent values.  Refer to subroutine readXsection_natural_mann_vertices in diffusive.f90 

3. In the same subroutine, curves of conveyance vs. elevation and dK/dA vs. elevation has been modified to make sure that the curves be monotonically increasing.  K here represents conveyance and A area.  The main reason of the modification is that Newton-Rapson uses first order derivative information for, in theory, quickly converging to its solution, which requires smooth and continuous functions.  Conveyance values especially in transition between in-bank and over-bank flows generally take nonlinearly behaving values, for example, sudden drop and then back up after gradually increasing in the respective flow regime.  This is mainly due to sudden increase of wetted perimeter when changing from in-bank to over-bank flow.

4. As existing water depth method using only Newton-Rapson method without solution bounds was experiencing abnormal results, a new hybrid method utilizing both N-R and Bisection methods within given solution bounds is built mainly in functions rtsafe and funcd_diffdepth in diffusive.f90.

5. Python files such as __main__.py, preprocess.py, compute.py, and diffusive_utils.py have been modified to pass topobathy data to the Fortran diffusive kernels.

## Additions
- Subroutine readXsection_natural_mann_vertices in diffusive.f90 to read natural cross sections. It also computes composite manning's N and uniform flow to be added to hydraulic lookup tables as an additional column.  The addition removes existing method of computing normal depth that uses an iterative numerical method, which can reduce compute time. 
- Functions rtsafe and  funcd_diffdepth in diffusive.f90 to compute water depth from computed flow values using a hybrid of Newton Rapson and Bisection methods.  The selection between the methods depend on the convergence rate as well as converging values of N-R.  When the rate is too slow or the values go outside of given solution bounds, then it switch to Bisection method.
-  Argument topobathy_data or topobathy_data_by_tw in __main__.py, preprocess.py, and compute.py is created to prepare  x_bathy_g, z_bathy_g, mann_bathy_g, and size_bathy_g from diffusive_utils.py. These four arguments are passed down to diffusive.f90.
- Two configuration variables in yaml file, that is, run_hybrid_routing and use_natl_xsections, are added.  These are boolean variables that indicate whether or not hybrid routing and natural cross sections are to be included in a simulation. Their values default to `False`, so setting them to `True` is mandatory for hybrid and natural cross section capabilities. Additionally, when `use_natl_xsections` is `False`, then trapezoidal geometries from the RouteLink file are used. (Thanks Adam for this description!!)

## Removals
- Newton-Rapson only method to compute water depth is removed and replaced by the new hybrid method. 

## Changes

-

### 1. Numerical stability test:
- Apply monotonically increasing curve for conveyance and its derivative wrt water area. Here, the next increasing point to be picked after a portion of curve decreasing is assumed to be 2~3% higher than a data point just before the drop.

image
image

- Apply the hybrid method to compute water depth. You can see clear improvement in terms of reducing oscillation of computed water depth value and abnormal jump of hydrograph in receding limbs.  Note that discharge value is affected by water depth value via both celerity and diffusivity that are both functions of water depth. That's why incorrect water depth values result in incorrect values of discharge.   

image
image

### 2. Routing applications to actual domain by Adam Wlostowski:
- Lower Colorado River, TX. 
- 1-month simulation, October 1 - November 15, 2019
- Diffusive domain is the final reach of the network: `[3766376, 3766378, 3766334]`
  - two tributary inflows: `[3766374, 3766372]`
- simulation yaml: `/glade/u/home/adamw/projects/t-route/test/jobs/lower_colorado_hybrid/natural_xsec_dev/colorado_natxs.yaml` 
- diffusive domain yaml: `/glade/u/home/adamw/projects/t-route/test/jobs/lower_colorado_hybrid/natural_xsec_dev/diffusive_domain.yaml`

Here, I am evaluating flow and depth results at the down-stream most segment of the diffusive domain, `3766334`. The natural cross section geometry shows that channel is deeply incised, with the channel bed lying ~21 meters below sea level. 
image

Hybrid simulations with and without natural cross sections were stable and produces similar looking flows:
image

As expected, water surface elevation profiles are drastically different across simulations. The simulation with natural cross section maintains a water surface elevation below sea level, due to the bed elevation of the cross section
image

In both diffusive hybrid simulations (with and without natural cross sections), I am noticing pretty strong differences in hydrograph shape compared to the Muskingum Cunge result. This is best seen to zooming in on a single event in the greater testing time domain: 
image
Notice how the diffusive wave has a very steep rising limb that arrives at a higher peak flow, compared to the MC simulation. Moreover, the falling limb falls rapidly, compared to the MC simulation. 

I'd like us to look into the behavior further to make sure it isn't related to some numerical instability. However, solving this issue is outside of the scope of this pull request. 

## Notes

-

## Todos

-

## Checklist

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


doc/v3_doc.yaml
```diff
@@ -390,6 +390,13 @@ compute_parameters:
         # of diffusive network segments, organized by tailwater ID (keys).
         # (!!) mandatory for hybrid simulations
         diffusive_domain:
+        # ---------------
+        # filepath to topobathy data for channel cross sections. 
+        # This file is currently only NetCDF file format and contain  
+        # topobathy data of a channel cross section defined by comid.
+        # Theoretically, comid is identical to link of RouteLink data.
+        # (!!) mandatory for diffusive routing for natural cross sections.
+        topobathy_domain:
```
awlostowski-noaa, line 399: Please add configuration documentation for `run_hybrid_routing` and `use_natl_xsections`. These are boolean variables that indicate whether or not hybrid routing and natural cross sections are to be included in a simulation. Their values default to `False`, so setting them to `True` is mandatory for hybrid and natural cross section capabilities. Additionally, when `use_natl_xsections` is `False`, then trapezoidal geometries from the RouteLink file are used

awlostowski-noaa (changes requested):
Just request that we update the documentation, otherwise this is good to go!

awlostowski-noaa (approved):
Thanks for making these changes. I'll go ahead and approve!

PR #517, opened 2022-02-08
Full Analysis and Assimilation test for the Lower Colorado, TX

awlostowski-noaa:
This is a canned analysis and assimilation test on the Lower Colorado River, TX with all necessary data. This test is intended to ship with the repository to allow quick start use on a test case that has nearly all capabilities operating within it. 

This test allows the user to experiment with the following t-route capabilities

- Muskingum Cunge channel routing
- Hybrid Muskingum Cunge - Diffusive Wave channel routing
- Level pool reservoir routing
- RFC reservoir forecasts
- Streamflow data assimilation

The only capability not included in this test is hybrid reservoir data assimilation. This domain simply does not include any of the USGS or USACE reservoirs for which reservoir data assimilation occurs in the National Water Model. 

## Additions

- A new directory `t-route/test/LowerColorado_TX` and sub-directories containing all requisite data necessary to execute the test
- A `.yaml` configuration file, `test_AnA.yaml` designed for a full analysis and assimilation simulation. 
- All data files have been subset so as to only include data for stream segments, lakes, gages, etc. that are within the Lower Colorado River basin. 

## Removals

- None

## Changes

- None

## Testing

1. `t-route/test/LowerColorado_TX/test_AnA.yaml`


jhrehanoaa (approved):
This runs correctly and successfully on my end.

jameshalgren:
This will not build with the comments in `$REPOROOT/src/fortran_routing/Reservoir_Binding/Reservoirs/makefile`

The `bind_hybrid.a` object is never moved to the lib folder for compile with the setup.py. 

If bind_hybrid is supposed to be deprecated, then setup.py in the framework folder needs to be modified to not refer to it. If the comments are not intentional, they need to be removed so that bind_hybrid.a is properly built and referenced. 

@karnesh

awlostowski-noaa:
@jameshalgren turns out bind_hybrid.a is no longer needed because we've re-wrote the hybrid reservoir da in pure python #509 

Interesting that this came up because the Lower Colorado domain does not have any USGS/ASCE reservoirs for which hybrid DA would be needed. However, if this is disrupting the build, then my point is moot. I will look into this now. 

FYI - considering a MASSIVE house cleaning and some repackaging with #519

jameshalgren:
And @awlostowski-noaa... kudos on #519 (It's a bit like having to go through one's box of school art drawings... couldn't we just keep one?)

PR #554, opened 2022-04-20
Woodchipper performance boost idea

awlostowski-noaa:
This is a first-cut exploration into re-configuring the ordering of network-scale MC calculations.


awlostowski-noaa:
@shorvath-noaa as you can see, the solution goes deep. I think I did most of the lower level modifications, we need to get it working with other reservoir modules etc.

jameshalgren (commented):
Seems like this could be implemented as an alternate compute option and tested side-by-side.

PR #556, opened 2022-04-21
MC kernel updates for V3.0 RouteLink

awlostowski-noaa:
The v3.0 RouteLink file contains stream segments with zero-valued floodplain widths, which causes issues in the existing MC kernel - specifically NaN flow predictions. The same issues are also caused in the WRF Hydro MC kernel - addressed in this PR: https://github.com/NCAR/wrf_hydro_nwm_public/pull/599

The changes proposed here include some additional conditions on if statements that account for situations when the water height is greater than the trapezoidal channel bank full height and the floodplain width is zero. 

## Testing

1. The newly changed MC kernel produces the same exact results as the old one. 
2. NCAR will conduct additional testing to make sure the changes work with the v3.0 RouteLink


awlostowski-noaa:
@aubreyd I tested @ CONUS scale with v 3.0 RouteLink and LAKEPARM, cold started. No nans to be found, so I'm optimistic that these changes work. Curious to see how your assessment turns out.

PR #563, opened 2022-05-24
updating refactor test config input

BrianAvant-NOAA:
Updating `diffusive_domain` path in `lower_colorado_refac.yaml` Lower Colorado refactor test case.


kumdonoaa (approved):
code change for new diffusive domain yaml file. Thanks.

PR #564, opened 2022-05-26
Remove mask filter for natural cross sections

BrianAvant-NOAA:
Removing mask filter for natural cross sections


kumdonoaa (commented):
Can you briefly describe what these changes are for?

PR #578, opened 2022-08-19
initial refactoring of main from ngen

shorvath-noaa:
Initial refactoring of main only incorporating edits from the ngen branch. Functionality has been added to current implementation rather than overwriting existing functions.

Next steps are to incorporate more recent changes in main branch __main__.py into ngen implementation. Start with addressing input parameters to compute_nhd_routing_v02 within _run_everything_v02 function in __main__.py

## Additions

- importing abstract networks in __main__.py
- added functions _handle_args_v02, main_v02, _run_everything_v02, to __main__.py
- added function _input_handler_v02 to input.py
- added function set_supernetwork_parameters to nhd_network_utilities_v02.py

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ ] PR has an informative and human-readable title
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code can be automatically merged (no conflicts)
- [ ] Code follows project standards (link if applicable)
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [ ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Target Environment support

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS
- [ ] Flexible from small to large screens
- [ ] No linting errors or warnings
- [ ] JavaScript tests are passing

PR #579, opened 2022-08-31
Da abstract classes

shorvath-noaa:
Updates to DataAssimlilation.py to create a data_assimilation object containing all the necessary DA information. This implementation defines a single DA class, 'AllDA', that creates all of the DA dataframes and dictionaries (e.g., lastobs_df, usgs_df, reservoir_usgs_df, etc.) that are needed based on user input. All of the logic regarding which DA options to use are performed within the AllDA class definition.

I've tested that calling AllDA(...) creates identical dataframes as those in the current implementation of t-route. I used the standard_AnA.yaml configuration file for these tests.

## Additions

- import additional necessary modules
- _reindex_link_to_lake_id function to replace link ID index values with lake ID
- _create_usgs_df function to create dataframe of USGS gage observations from USGS timeslice files
- _create_reservoir_df function which is generalized to create either usgs or usace reservoir dataframes based on input
- _set_reservoir_da_params function to update reservoir parameters from run_results
- new_lastobs function to create new lastobs_df for next simulation chunk
- AllDA class which is a subclass of DataAssimilation. Determines which DA features to turn on based on user input and creates all necessary dataframes
- update function within AllDA that updates dataframes for next simulation chunk

## Removals

-

## Changes

-

## Testing

1. Once NHDNetwork object is updated, test to make sure these two are compatible. AllDA takes 'network' as an input, so we need to make sure network has the necessary features to run AllDA.

## Notes

-

## Todos

- We might consider creating separate DA subclasses for each type of DA (streamflow nudging, reservoir persistence, etc.). Right now it is all done in a single DA subclass, 'AllDA'

## Checklist

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/src/troute-network/troute/DataAssimilation.py
+++ b/src/troute-network/troute/DataAssimilation.py
@@ -15,6 +15,35 @@ def build_data_assimilation_csv(data_assimilation_parameters):
         data_assimilation_parameters["wrf_hydro_da_channel_ID_crosswalk_file"],
     )
 
+def build_data_assimilation_folder_v02(data_assimilation_parameters, run_parameters, network):
+
+    usgs_timeslices_folder = pathlib.Path(
+        data_assimilation_parameters["data_assimilation_timeslices_folder"],
+    ).resolve()
+    if "data_assimilation_filter" in data_assimilation_parameters:
+        da_glob_filter = data_assimilation_parameters["data_assimilation_filter"]
+        usgs_files = sorted(usgs_timeslices_folder.glob(da_glob_filter))
+    elif "usgs_timeslice_files" in data_assimilation_parameters:
+        usgs_files = data_assimilation_parameters.get("usgs_timeslice_files", None)
+        usgs_files = [usgs_timeslices_folder.joinpath(f) for f in usgs_files]
+    else:
+        print("No Files Found for DA")
+        # TODO: Handle this with a real exception
+    
+    return nhd_io.get_obs_from_timeslices(
+        network.link_gage_df,
+        # next two lines assume we've already verified that streamflow_da is true and contains the necessary dictionary items.
+        #TODO: reorganize how DA objects are called/created in __main__.py to ensure we have all necessary info...
+        data_assimilation_parameters.get('streamflow_da',None).get('crosswalk_gage_field','gages')
+        data_assimilation_parameters.get('streamflow_da',None).get('crosswalk_segID_field','link'),
+        usgs_files,
+        data_assimilation_parameters.get("qc_threshold", 1),
+        data_assimilation_parameters.get("data_assimilation_interpolation_limit", 59),
+        900, # 15 minutes, as secs
+        run_parameters["t0"],
+        network.cpu_pool, #verify
+    )
+
 def build_data_assimilation_folder(data_assimilation_parameters, run_parameters):
 
     usgs_timeslices_folder = pathlib.Path(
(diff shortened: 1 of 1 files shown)


shorvath-noaa:
@kumdonoaa @hellkite500 @mattw-nws : The only changes are to the DataAssimilation.py file, so t-route will run as normal with these changes. This new object will be implemented once we refactor __main__.py which has a draft PR out (https://github.com/NOAA-OWP/t-route/pull/578)

src/troute-network/troute/DataAssimilation.py
```diff
@@ -115,4 +386,438 @@ def last_obs(self):
...
+        - da_parameter_dict             (dict): user input data re data assimilation
+        - waterbody_types_df       (DataFrame): Waterbody type codes (1:levelpool, 2:USGS, 3:USACE, 4:RFC)
+        - usgs_lake_gage_crosswalk      (dict): usgs gage ids and corresponding segment 
+                                                ids at which they are located
+        - usace_lake_gage_crossalk      (dict): usace gage ids and corresponding segment 
+                                                ids at which they are located
+    """
+    __slots__ = ["_usgs_df", "_last_obs_df", "_reservoir_usgs_df", "_reservoir_usgs_param_df", "_reservoir_usace_df", "_reservoir_usace_param_df", "_da_parameter_dict", "_waterbody_types_df", "_usgs_lake_gage_crosswalk", "_usace_lake_gage_crosswalk"]
+    def __init__(self, data_assimilation_parameters, run_parameters, waterbody_parameters, network, da_run):
+        lastobs_df, da_parameter_dict = nnu.build_data_assimilation_lastobs(
```
hellkite500, line 426: why isn't this `build_data_assimilation_lastobs` functions moved to this module?

src/troute-network/troute/DataAssimilation.py
```diff
@@ -115,4 +386,438 @@ def last_obs(self):
...
+            rfc_forecast = False
+        
+        level_pool_params = waterbody_parameters.get('level_pool', defaultdict(list))
+                
+                
+        #--------------------------------------------------------------------------------
+        # Assemble USGS observation data for Streamflow DA or USGS Reservoir Persistence
+        #--------------------------------------------------------------------------------
+
+        # if user requested nudging or usgs_persistence and a specified a USGS TimeSlice directory,
```
hellkite500, line 480: slightly confusing comment "and a specified a USGS TimeSlice directory"

src/troute-network/troute/DataAssimilation.py
```diff
@@ -115,4 +386,438 @@ def last_obs(self):
...
+            break_network_at_waterbodies = False
+        
+        if break_network_at_waterbodies:
+            if (param_file and (usgs_persistence or usace_persistence)) or (param_file and rfc_forecast):
+                waterbody_type_specified = True
+                (
+                    waterbody_types_df, 
+                    usgs_lake_gage_crosswalk, 
+                    usace_lake_gage_crosswalk
+                ) = nhd_io.read_reservoir_parameter_file(
```
hellkite500, line 503: Hmm...here is a strong connection of DA back to the nhd network, which isn't what we want with these classes.  We may need to take a close look at what the purpose of this is and where the functionality is best suited to exist.

hellkite500 (commented):
Definitely going in the right direction, but we may need to discuss the decoupling of network and DA functionality a bit...

src/troute-network/troute/DataAssimilation.py
shorvath-noaa: dropped the 'a' before 'specified'. I think that makes it clearer?

src/troute-network/troute/DataAssimilation.py
shorvath-noaa: No reason. I've now included this function within this module.

src/troute-network/troute/DataAssimilation.py
shorvath-noaa: This function just relates gage ids to reservoir ids, so I believe this is still used for ngen, no? I don't think this has anything to do with the type of network.... But I will move this function into this module so we are not relying on nhd_io.py

hellkite500: What I meant here is that the way this function works, it relies on some state/info in a reservoir param file that is specific to a particular network representation (in this case, the nhd/nwm reservoir params).  What I was suggesting is refactoring this so that DA can extract that information from an instantiated Network (whether it is an `NHDNetwork`, `HYFeaturesNetwork`, `CustomNetwork`, ect...) instead of reading hard coded params from a file that require all networks to use the same file format/layout ect...

shorvath-noaa:
Changes have been made to this module so that it should be network-agnostic now.

kumdonoaa (approved):

PR #580, opened 2022-08-31
update of NHDNetwork class for ngen t-route

kumdonoaa:
[Short description explaining the high-level reason for the pull request]
As part of goal for integrating t-route model developed for NMW v3.0 into Next Gen Water Modeling Framework, NHDNetwork class is modified  to produce needed network and input forcing data for running t-route through BMI.  t-route with these data produced the exactly same results as t-route from t-route main branch. 
## Additions
NHDNetwork_v4.py is used instead of NHDNetwork.py only for test purpose.

-

## Removals

-

## Changes
AbstractNetwork.py for adding _coastal_boundary_depth as an abstract property
__main__.py for first instantiating NHDNetwork class so as to prepare network and forcing data and then for running t-route with such prepared data.
-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/src/troute-network/troute/NHDNetwork.py
+++ b/src/troute-network/troute/NHDNetwork.py
@@ -1,4 +1,4 @@
-from  .AbstractNetwork import AbstractNetwork
+ from  .AbstractNetwork import AbstractNetwork
 import xarray as xr
 import pathlib
 from collections import defaultdict
--- a/src/troute-network/troute/NHDNetwork_v4.py
+++ b/src/troute-network/troute/NHDNetwork_v4.py
@@ -1,0 +1,318 @@
+from  .AbstractNetwork import AbstractNetwork
+import xarray as xr
+import pathlib
+from collections import defaultdict
+import troute.nhd_io as nhd_io
+import pandas as pd
+import time
+
+from nwm_routing.preprocess import (
+    nwm_network_preprocess,
+    nwm_initial_warmstate_preprocess,
+    nwm_forcing_preprocess,
+    unpack_nwm_preprocess_data,
+    )
+import troute.nhd_network_utilities_v02 as nnu
+import logging
+LOG = logging.getLogger('')
+
+__showtiming__ = True #FIXME pass flag
+__verbose__ = True #FIXME pass verbosity
+
+def read_qlats(forcing_parameters, segment_index):
+    # STEP 5: Read (or set) QLateral Inputs
+    if __showtiming__:
+        start_time = time.time()
+    if __verbose__:
+        print("creating qlateral array ...")
+    qts_subdivisions = forcing_parameters.get("qts_subdivisions", 1)
+    nts = forcing_parameters.get("nts", 1)
+    qlat_input_folder = forcing_parameters.get("qlat_input_folder", None)
+    qlat_input_file = forcing_parameters.get("qlat_input_file", None)
+    if qlat_input_folder:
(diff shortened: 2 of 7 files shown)


src/troute-network/troute/NHDNetwork.py
```diff
@@ -1,4 +1,4 @@
-from  .AbstractNetwork import AbstractNetwork
+ from  .AbstractNetwork import AbstractNetwork
```
hellkite500, line 1: This extra space at the beginning of the line may cause some issues...either way, for consistency it should be removed.

src/troute-network/troute/NHDNetwork_v4.py
```diff
@@ -0,0 +1,318 @@
+from  .AbstractNetwork import AbstractNetwork
```
hellkite500, line 1: This file should simply replace the existing NHDNetwork.py file.  This way the changes from one version to the next are marked in the version control.  A well crafted commit message for this single change will make going back into the history and finding the old version straight forward, and we don't have multiple files to maintain.

src/troute-nwm/src/nwm_routing/__main__.py
```diff
@@ -28,6 +33,689 @@
 
 LOG = logging.getLogger('')
 
+'''
```
hellkite500, line 36: we really don't need to maintain v02 main, as long as the functionality of v04 main works as needed.  I would propose dropping v02 main and make sure v04 (which should probably just be `main`) is functional.

src/troute-nwm/src/nwm_routing/__main__.py
```diff
@@ -714,6 +1403,322 @@ def main_v03(argv):
...
+            output_end_time = time.time()
+            task_times['output_time'] += output_end_time - ic_end_time
+            
+    if showtiming:
+        task_times['total_time'] = time.time() - main_start_time
+
+    LOG.debug("process complete in %s seconds." % (time.time() - main_start_time))
+    
+    if showtiming:
+        print('************ TIMING SUMMARY ************')
```
hellkite500, line 1672: Since task_times contains all the relevant info, I would pull this logic into a function and pass in `task_times`

src/troute-nwm/src/nwm_routing/__main__.py
```diff
@@ -1034,11 +2039,16 @@ async def main_v03_async(argv):
...
+    #if v_args[0].input_version == 4:
+    #    LOG.info("Running main v03 - async looping")
+    #    coroutine = main_v03_async(v_args[1])
+    #    asyncio.run(coroutine)
         # loop.run_until_complete(coroutine)
+
     if v_args[0].input_version == 3:
         LOG.info("Running main v03 - looping")
         main_v03(v_args[1])
+    if v_args[0].input_version == 4:
```
hellkite500, line 2051: Again, let's try to get to a single main function.  If backwards compatibility is important, we can discuss ways to support that, but with 4 versions of main and an async v3 interface, it is reasonably confusing.

hellkite500 (changes requested):
A couple things we really need to address before this moves forward further.  Please see the comments regarding the internal file versioning semantics as well as the file name versioning.

PR #597, opened 2022-12-08
Raise error with unknown extensions.

groutr:
Raise some sort of error when an unknown file extension is passed to read_file. This will avoid a unbound variable.


kumdonoaa (approved):
Great to always have this kind of checkpoint.

groutr:
The `building_forcing_sets` in this same file function also has several issues. One thing I'm perplexed by is the assumption of a very specifically formatted glob string on L117 (https://github.com/NOAA-OWP/t-route/blob/master/src/troute-network/troute/hyfeature_network_utilities.py#L117).

PR #629, opened 2023-08-04
Time log added to troute_model

AminTorabi-NOAA:
Time log is added to troute_model.py to record the time take for creation of network, forcing , data assimilation, routing and  output time.

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/src/troute-network/troute/HYFeaturesNetwork.py
+++ b/src/troute-network/troute/HYFeaturesNetwork.py
@@ -431,9 +431,10 @@ class HYFeaturesNetwork(AbstractNetwork):
             gages_df[['type','value']] = gages_df.hl_uri.str.split('-',expand=True,n=1)
             # filter for 'Gages' only
             gages_df = gages_df[gages_df['type'].isin(['Gages','NID'])]
-            # assign only the furthest downstream segment ID to any given gage
+            # In the event that a segment contains multiple gages, assign only the furthest 
+            # downstream segment ID to any given gage
             gages_df = gages_df.sort_values(['value','hydroseq']).groupby('value').last().reset_index()
-            # transform dataframe into a dictionary
+            # transform dataframe into a dictionary where key is segment ID and value is gage ID
             self._gages = gages_df[['id','value']].rename(columns={'value': 'gages'}).set_index('id').to_dict()
 
             # Create lake_gage crosswalk dataframes:


src/troute_model.py
```diff
@@ -27,6 +28,15 @@ def __init__(self, bmi_cfg_file):
...
         
+        self.task_times = {                                     # Creating a dictionary to record time for each section #
+            'network_time': 0,
+            'forcing_time' : 0,
+            'output_time' : 0,
+            'run_time' :0,
+            'data_assimilation_time' : 0,
+
+        }
+        self.main_start_time = time.time()
```
shorvath-noaa, line 39: main_start_time is called here but never referenced again.

Because t-route is being called via BMI functions and other formulations will also be running, we'll need to rethink how we log the total time. Calling time.time() here and then again after all is said and done would actually record the timing of t-route PLUS all of the other formulations. I think if we cleverly place all the other timing calls we can simply add them all up when we finalize() the BMI instance of t-route.

src/troute_model.py
```diff
@@ -78,7 +91,20 @@ def preprocess_static_vars(self, values: dict):
...
+                del self._network._reverse_network[key]
+                for tw in self._network._independent_networks.keys():
+                    del self._network._independent_networks[tw][key]
+                    for rli, _ in enumerate(self._network._reaches_by_tw[tw]):
+                        self._network._reaches_by_tw[tw][rli].remove(key)
+
+        network_end_time = time.time()                          # end of recording time for network creation #
+        self.task_times['network_time'] += network_end_time - network_start_time
 
+        DA_start_time = time.time()                             # start of recording time for data assimilation creation #
```
shorvath-noaa, line 107: We call time.time() twice here basically in a row. We could rewrite this to only call it once and use that same time call for both network_time and data_assimilation_time

src/troute_model.py
```diff
@@ -127,15 +148,24 @@ def run(self, values: dict, until=300):
         else:
             flowveldepth_interorder = {}
 
+        forcing_end_time = time.time()                          # end of recording time for forcing values #
+        self.task_times['forcing_time'] += forcing_end_time - forcing_start_time
+        
         # Data Assimilation values:
+        DA_start_time = time.time()                             # start of recording time for data assimilation creation #
+
```
shorvath-noaa, line 156: see previous comment about combining our usage of time.time()

src/troute_model.py
```diff
@@ -199,13 +229,16 @@ def run(self, values: dict, until=300):
...
         self._network.update_waterbody_water_elevation()               
-        
+        run_end_time = time.time()                              # end of recording time for routing #
+        self.task_times['run_time'] += run_end_time - run_start_time
+
         # update t0
         self._network.new_t0(self._time_step, nts)
 
         # get reservoir DA initial parameters for next loop iteration
         self._data_assimilation.update_after_compute(self._run_results)
```
shorvath-noaa, line 239: For now let's include these in the run_time we end on line 232.

shorvath-noaa:
@AminTorabi-NOAA Let's implement if: statements to only log timings if it is specified in the config file. See how this is done in [__main__.py](https://github.com/NOAA-OWP/t-route/blob/f3d5b1b1fcc2de8d80781dfa118a7777f70fd4d5/src/troute-nwm/src/nwm_routing/__main__.py#L61-L62). We'll need to also add 'log_parameters' to [troute_model.py](https://github.com/NOAA-OWP/t-route/blob/master/src/troute_model.py) that are read in when [_read_config_file()](https://github.com/NOAA-OWP/t-route/blob/f3d5b1b1fcc2de8d80781dfa118a7777f70fd4d5/src/troute_model.py#L236) is called.

src/troute_model.py
AminTorabi-NOAA: Self.main_start_time can be used to log total time if needed. But I don't think its necessary here since we breaking it down. I'll delete it.

src/troute_model.py
AminTorabi-NOAA: Great point. Basically DA_start_time = network_end_time
I'll delete line 107 and use network_end_time

Issue #658, opened 2023-10-14
Error in running


TrupeshKumarPatel:
After successful installation, I tried to run a few test cases, and none were running.

## Current behavior

One of the test cases I tried is [test/LowerColorado_TX](https://github.com/NOAA-OWP/t-route/blob/master/test/LowerColorado_TX/test_AnA.yaml).

## Error

```python3
Traceback (most recent call last):
  File "/usr/lib64/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib64/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.9/site-packages/nwm_routing/__main__.py", line 2028, in <module>
    main_v04(v_args[1])
  File "/usr/local/lib/python3.9/site-packages/nwm_routing/__main__.py", line 54, in main_v04
    ) = _input_handler_v04(args)
  File "/usr/local/lib/python3.9/site-packages/nwm_routing/input.py", line 44, in _input_handler_v04
    troute_configuration = Config(**data)
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 6 validation errors for Config
network_topology_parameters -> supernetwork_parameters -> geo_file_type
  extra fields not permitted (type=value_error.extra)
network_topology_parameters -> waterbody_parameters -> rfc
  extra fields not permitted (type=value_error.extra)
compute_parameters -> data_assimilation_parameters -> streamflow_da -> wrf_hydro_lastobs_file
  extra fields not permitted (type=value_error.extra)
compute_parameters -> data_assimilation_parameters -> reservoir_da -> gage_lakeID_crosswalk_file
  extra fields not permitted (type=value_error.extra)
compute_parameters -> data_assimilation_parameters -> reservoir_da -> reservoir_persistence_usace
  extra fields not permitted (type=value_error.extra)
compute_parameters -> data_assimilation_parameters -> reservoir_da -> reservoir_persistence_usgs
  extra fields not permitted (type=value_error.extra)
```

[...]

Issue #651, opened 2023-10-05
Error in installation 


TrupeshKumarPatel:
I tried to follow the readme installation instructions and build a t-route in the system but got the following error.

## OS Information
_OS Name_: `Rocky Linux`
_OS Version_: `9.2`

## Pre-installed Packages
```bash 
dnf install tar git gcc-c++ gcc gcc-gfortran  make cmake bzip2 python3 python3-devel python3-pip udunits2-devel -y
dnf install libstdc++ libstdc++-devel glibc glibc-devel libgfortran openmpi openmpi-devel -y
dnf install lapack gdal-libs gdal gdal-devel -y
dnf install netcdf netcdf-devel netcdf-fortran netcdf-fortran-devel netcdf-openmpi netcdf-openmpi-devel netcdf-fortran-openmpi netcdf-fortran-openmpi-devel -y 
pip3 install wheel deprecated dask pyarrow geopandas pyproj fiona pandas xarray netcdf4 joblib toolz pyyaml Cython
```

## Installation Steps
```bash 
git clone --progress --single-branch --branch master http://github.com/NOAA-OWP/t-route.git
cd t-route
./compiler.sh
```

## Compilation Log
```bash
using F90=gfortran
using CC=gcc
using NETCDFINC=/usr/include
rm -f *.o *.mod
gfortran -g -c -O2 -fPIC -lgfortran -lgcc -static-libgfortran -static-libgcc -nodefaultlibs varPrecision.f90
gfortran -g -c -O2 -fPIC -lgfortran -lgcc -static-libgfortran -static-libgcc -nodefaultlibs -o mc_single_seg.o MCsingleSegStime_f2py_NOLOOP.f90
gfortran -g -c -O2 -fPIC -lgfortran -lgcc -static-libgfortran -static-libgcc -nodefaultlibs -o pymc_single_seg.o pyMCsingleSegStime_NoLoop.f90
cp *.o ../../../src/troute-routing/troute/routing/fast_reach
rm -f *.o *.mod
gfortran -g -c -O2 -fPIC -fbounds-check -o diffusive.o diffusive.f90
diffusive.f90:1440:20:

 1440 |          pause !test
      |                    1
Warning: Deleted feature: PAUSE statement at (1)
gfortran -g -c -O2 -fPIC -fbounds-check -o pydiffusive.o pydiffusive.f90
cp *.o ../../../src/troute-routing/troute/routing/fast_reach
rm -f *.o
rm -f *.mod
[...]

Issue #617, opened 2023-06-26
test_AnA test case fails with KeyError: "['index'] not found in axis"


aaraney:
I followed the readme's Usage and Testing section to install t-route plus installed the python packages `geopandas`, `deprecated`, `pyarrow` manually. Post install, I attempted to run the `test_AnA` test case but ran into a `KeyError: "['index'] not found in axis"` error during the preprocessing phase of the test. See the below output:

<details>
<summary>Shell output</summary>

```shell
python -m nwm_routing -f -V3 test_AnA.yaml
2023-06-26 19:50:28,116   DEBUG [input.py:199 -         check_inputs()]: ***** Begining configuration file (.YAML) check *****
2023-06-26 19:50:28,116   DEBUG [input.py:105 -     _does_file_exist()]: Checking that a geo_file_path parameter is specified and the file exists
2023-06-26 19:50:28,116   DEBUG [input.py:136 -     _does_file_exist()]: The geo_file_path parameter is specified and exists
2023-06-26 19:50:28,116   DEBUG [input.py:105 -     _does_file_exist()]: Checking that a mask_file_path parameter is specified and the file exists
2023-06-26 19:50:28,116   DEBUG [input.py:136 -     _does_file_exist()]: The mask_file_path parameter is specified and exists
2023-06-26 19:50:28,116   DEBUG [input.py:105 -     _does_file_exist()]: Checking that a level_pool_waterbody_parameter_file_path parameter is specified and the file exists
2023-06-26 19:50:28,116   DEBUG [input.py:136 -     _does_file_exist()]: The level_pool_waterbody_parameter_file_path parameter is specified and exists
2023-06-26 19:50:28,116   DEBUG [input.py:105 -     _does_file_exist()]: Checking that a wrf_hydro_channel_restart_file parameter is specified and the file exists
2023-06-26 19:50:28,116   DEBUG [input.py:136 -     _does_file_exist()]: The wrf_hydro_channel_restart_file parameter is specified and exists
2023-06-26 19:50:28,116   DEBUG [input.py:105 -     _does_file_exist()]: Checking that a wrf_hydro_waterbody_ID_crosswalk_file parameter is specified and the file exists
2023-06-26 19:50:28,116   DEBUG [input.py:136 -     _does_file_exist()]: The wrf_hydro_waterbody_ID_crosswalk_file parameter is specified and exists
[...]

PR #660, opened 2023-10-21
Fixed error and Added Workflow to trigger GitHub Action for `master` branch

TrupeshKumarPatel:
This PR does two tasks:

1. Solves Installation problem #651 and test run issues (#658, #617)
2. Added workflow to check the build and test the new PR on the `master` branch

## Additions

- Added a specific version of Cython to fix the Installation problem (suggested by [Joshua Cunningham](https://github.com/NOAA-OWP/t-route/issues/651#issuecomment-1772868472))
- Added if condition to drop the `index` column to fix runtime error (suggested by [James H.](https://github.com/JurgenZach-NOAA/t-route-jz1/pull/1#issue-1955139229))
- Added a python application workflow

## Changes

- [requirements.txt](https://github.com/TrupeshKumarPatel/t-route/commit/5112c77b3c1e06beb12a6ae0281f48dee13d9236#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552)
- [src/troute-nwm/src/nwm_routing/preprocess.py](https://github.com/TrupeshKumarPatel/t-route/commit/5112c77b3c1e06beb12a6ae0281f48dee13d9236#diff-774dbfdc53d8120df4a70b6a20c60e55433cb0f5080a22ff3d6da3d40542cb6d)
- [.github/workflows/python-app.yml](https://github.com/NOAA-OWP/t-route/pull/660/files#diff-d7b3e1cc0b909c52439e5d8d3f602f70644d0f6b62a54825968d819f7a5bee89)

## Testing

1. [test/LowerColorado_TX](https://github.com/NOAA-OWP/t-route/blob/master/test/LowerColorado_TX/test_AnA.yaml)
```bash
cd test/LowerColorado_TX
python -m nwm_routing -f -V3 test_AnA.yaml
```

## Checklist

- [X] PR has an informative and human-readable title
- [X] Code can be automatically merged (no conflicts)
- [X] Code follows project standards (link if applicable)

- [ ] Windows
- [X] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


src/troute-nwm/src/nwm_routing/preprocess.py
```diff
@@ -1094,10 +1096,12 @@ def nwm_forcing_preprocess(
...
             usgs_df_15min.join(link_lake_df, how = 'inner').
             reset_index().
-            set_index('usgs_lake_id').
-            drop(['index'], axis = 1)
+            set_index('usgs_lake_id')
         )
         
+        if 'index' in reservoir_usgs_df.columns:
+            reservoir_usgs_df=reservoir_usgs_df.drop(['index'], axis = 1)
+
```
shorvath-noaa, line 1104: I think an alternative solution here would be to use the `drop=` parameter in `.reset_index()`. So:

```
        link_lake_df = (
            gage_lake_df.
            join(gage_link_df, how = 'inner').
            reset_index(drop=True).set_index('link')
        )
```

I'm not sure the cause of this issue (pandas version maybe?), but what was happening is that `reset_index()` makes the existing index a new column in the dataframe, sometimes naming it 'index', sometimes not. But either way, we want to remove that index all together, so using `reset_index(drop=True)` will do this regardless of the index name.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 numpy
-Cython
+Cython<=3.0.3
```
shorvath-noaa, line 2: Could you clarify why this version dependency?

src/troute-nwm/src/nwm_routing/preprocess.py
jameshalgren: @shorvath-noaa This sounds like a better, more comprehensive fix. 
@TrupeshKumarPatel, @JoshCu -- suggest we split these fixes into three PRs -- @shorvath-noaa, could you confirm that this would help?

shorvath-noaa: @jameshalgren @TrupeshKumarPatel [@JoshCU](https://github.com/JoshCu) yeah, I think splitting this up into separate PRs would help us test/approve these quickly.

JoshCu: While we're modifying requirements.txt, should we add 
1. deprecated
2.  pyarrow
3.  geopandas?

They're currently being manually installed in a bunch of docker images and the [ngen documentation](https://github.com/NOAA-OWP/ngen/blob/master/doc/PYTHON_ROUTING.md#setup-virtual-environment) too.

Wheel is also needed for building the wheels but I don't know if it's strictly necessary to run t-troute. Might be a small boost to the install speed of some of the other required packages though.

JoshCu: I've got some CI for building and testing a t-route docker image , but it needs the Lower/Colorado fix. Once it's merged I'll open a PR for the docker image

shorvath-noaa: I'd say yes to updating the requirements.txt. And I'm not too familiar with docker images, but we are working on getting a working version of LowerColorado_TX with nextgen, so I'll update you once that's ready.

TrupeshKumarPatel: @shorvath-noaa, the only reason I put the `if` condition is because, as per the previous code, the first `index` dropping was applied on `link_lake_df` DataFrame and then re-applied `index` dropping after joining operation. Now, because the `index` column was already dropped, it couldn’t find the next join. Therefore, if you do `reset_index(drop=True)`, you will get the same error. 

So, I decided to add `if` conditions on both DataFrame to avoid error in any case.

TrupeshKumarPatel:
Thank you, @JoshCu, for separating this PR.

shorvath-noaa:
@TrupeshKumarPatel I believe this PR can be closed now as it has been merged in separate PRs.

PR #666, opened 2023-10-27
Bmi compliant arrays

JurgenZach-NOAA:
Added routines to encode and decode BMI compliant arrays (flattened, and appropriately preprocessed time- and station ID axes). Encoding is implemented for USGS and USACE preprocessed time domain files in model_DAforcing. py. 

For the time being (to verify proper function), the reverse de-coding is also implemented in model_DAforcing.py. Said decoding will, in one of the next steps, be moved to DataAssimilation. However, the file can be checked in, b/c the decoiding can be easily disabled by setting the flag testRevert = False.

Next steps will be added to the same pull request. 

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


shorvath-noaa:
@JurgenZach-NOAA it looks like this PR also includes the newly created forcing files for LowerColorado_TX_V4. Are these necessary for this PR? If not, could you remove these?

JurgenZach-NOAA:
Sean: they were not needed, and I removed them, thanks for pointing out!

JurgenZach-NOAA:
Added rfc encoding & decoding.

JurgenZach-NOAA:
Cleaned up file from debug statements.

src/model_DAforcing.py
```diff
@@ -117,9 +118,472 @@ def __init__(self, bmi_cfg_file=None):
...
+                # Decode station ID axis
+                stationAxisName = 'stationId'
+                df_withStationsAndDates_reservoirUsgs = _stations_retrieve_from_arrays(df_withDates_reservoirUsgs, \
+                    self.stationArray_reservoir_usgs, self.stationStringLengthArray_reservoir_usgs, \
+                    stationAxisName)
+                
+                # Reservoir USACE dataframe
+
+                # Unflatten the arrays
+                df_raw_reservoirUsace = _unflatten_array(self.reservoirUsaceArray,self.nDates_reservoir_usace,\
```
shorvath-noaa, line 279: Got the following error when running with USACE persistence set to `False`:

`AttributeError: 'DAforcing_model' object has no attribute 'reservoirUsaceArray'`

Looks like you need to make sure you set some of these slots to empty (empty dataframe, False, None, etc.) values if the DA feature is turned off.

shorvath-noaa (approved):

JurgenZach-NOAA (commented):
Reviewed for accuracy

src/model_DAforcing.py
```diff
@@ -175,6 +375,285 @@ def run(self, values: dict):
...
+    stationArray = np.empty(stationStringLengthArray[0], dtype=np.int64)
+
+    # init some aux. variables
+    firstIter = True
+    stationIndex = 0
+    # iterate through all stations
+    for station in stations:
+        # capture string length of present station
+        stationStringLengthArray[stationIndex] = len(station)
+        stationIndex += 1
```
hellkite500, line 410: I'll note here that BMI 2.0 doesn't explicitly allow passing arbitrary string type variables. This function may still be useful in the future with changes or extensions of BMI, but there will need to be some extra care in handling the string buffers.

src/model_DAforcing.py
```diff
@@ -123,6 +126,201 @@ def __init__(self, bmi_cfg_file=None):
...
+                #
+                # Dates related:
+                # - datesSecondsArray_reservoir_usgs
+                # - nDates_usgs
+                # 
+                # Station ID related:
+                # - stationArray_usgs 
+                # - stationStringLengthArray_usgs 
+                # - nStations_usgs 
+                #
```
hellkite500, line 182: This class seems to be getting a lot of attributes that are grouped/related. I strongly recommend creating new data structures/classes for these groups and then holding an instance in the class. This could also provide the basis for an abstract interface that allows new DA sources to more easily plug in in the future.

src/model_DAforcing.py
JurgenZach-NOAA: Just to clarify, none of the arrays contain strings (or characters) - sorry if the variable naming was misleading. Entries in dataframes are converted to integers by encoding into ASCII. The entire dataframe segment (e.g., column of station IDs) is converted into one single integer array, simply concatenating all the ASCII array representations of each station IDs together. Since we cannot always rely on each string (station ID) to have the same character length, a second array is created with the string lengths of the station IDs (or other entries such as filenames, etc). 

The specific array you flagged is actually said array of string lengths. 

So: the arrays as implemented are definitely BMI compliant.

src/model_DAforcing.py
JurgenZach-NOAA: That point is well taken, although the treatment of the dataframes is not as "universal" as it may appear at first. For example, the usgs- and usace dataframes (and, by extension, any future additional gauges included) are indeed similar (once preprocessed), but the rfc arrays, for example,, are quite different. 

However, once all the dataframes have been outfitted with a BMI transport option (a few weeks at most), we'll definitely evaluate opportunities to create new classes. One candidate is definitely said preprocessed usgs- and usace arrays (vs station IDs and date/time).

src/model_DAforcing.py
hellkite500: Hmm, I can see a possible path here...
The problem, I think, lies in this function in BMI 2.0:
https://bmi.readthedocs.io/en/stable/#get-var-itemsize
It requires any BMI variable (which may be an array) to say how many bytes a SINGLE element of that array contains (the entire array would have a total size given by https://bmi.readthedocs.io/en/stable/#get-var-nbytes).
So if you are ASCII encoding variable length strings into an array, you will need to pad each string to some maximum size.  Also should probably report the units of that variable as "ASCII" to indicate the encoding used.

We could extend the typically supported types to include "cstring"  in this way.

Unless the intent is to pass these as `char`, or unsigned short `int` arrays, which I think would work as is, but will require some extra protocol to decode.

Thanks for explaining a bit of this rational!

JurgenZach-NOAA: To clarify further: each entry of the array is simply an ASCII code, so indeed, short unsigned int would suffice. So, a set of entries in a dataframe that are, e.g., station IDs are encoded as TWO integer arrays. So, e.g., the two following strings: 

['S1', '007'] are encoded as:

stringArray: 
        ([ASCII code of 'S', ASCII code of '1', ASCII code of '0', ASCII code of '0', ASCII code of '7']), 
or:   
        ([83, 49, 48, 48, 55 ])

stringLengthArray:
        ([Length of 'S1'], [Length of '007']) , 
or:    ([2, 3]). 

Both encoding and decoding are implemented based on that.

JurgenZach-NOAA: Each item of the array is well defined, and can be represented as the same size integer. The point that we could use unsigned short integers is well taken, and we can definitely take a note for the next PR: will that also comply with any non-python components planned?

hellkite500: I'm thinking "out loud" on this thread... Moving arbitrarily sized strings (character/byte arrays) via BMI isn't something heavily explored, and isn't currently supported in the ngen runtime, so thinking about how that would be done. A fixed size array with `'\0'` terminated strings could avoid the need for a separate length array, but that depends on what the `"S1"` semantics are in the context of the identifier.

PR #683, opened 2023-11-07
function to save output with different intervals

AminTorabi-NOAA:
A function added to the DAforcing that can save output with different intervals.

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/src/model_DAforcing.py
+++ b/src/model_DAforcing.py
@@ -6,7 +6,7 @@ from joblib import delayed, Parallel
 import xarray as xr
 import glob
 import pathlib
-
+import netCDF4
 from troute.routing.fast_reach.reservoir_RFC_da import _validate_RFC_data
 
 from nwm_routing.log_level_set import log_level_set
@@ -354,13 +354,16 @@ class DAforcing_model():
         """
         output_parameters = self._output_parameters
         da_parameters = self._data_assimilation_parameters
-
-        if output_parameters['lite_restart'] is not None:
-            _write_lite_restart(
-                values,
-                self._t0,
-                output_parameters['lite_restart']
-            )   
+        q0 = values.get('q0')
+        waterbody_df = values.get('waterbody_df')
+        lastobs_df = values.get('lastobs_df')
+        if len(q0) != 0 and len(waterbody_df) != 0 and len(lastobs_df) != 0:
+            if output_parameters['lite_restart'] is not None:
+                _write_lite_restart(
+                    values,
+                    self._t0,
+                    output_parameters['lite_restart']
+                )   
 
         lastobs_output_folder = da_parameters.get('streamflow_da',{}).get('lastobs_output_folder', None)
         if lastobs_output_folder:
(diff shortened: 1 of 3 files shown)


src/model_DAforcing.py
```diff
@@ -354,13 +354,16 @@ def run(self, values: dict):
...
-        if output_parameters['lite_restart'] is not None:
-            _write_lite_restart(
-                values,
-                self._t0,
-                output_parameters['lite_restart']
-            )   
+        q0 = values.get('q0')
+        waterbody_df = values.get('waterbody_df')
+        lastobs_df = values.get('lastobs_df')
+        if len(q0) != 0 and len(waterbody_df) != 0 and len(lastobs_df) != 0:
```
shorvath-noaa, line 360: I don't follow what this `if` statement is for? `q0` here should never be empty, it's the main output from t-route. `waterbody_df` could be empty if waterbodies are not being simulated, but `_write_lite_restart` already checks for this before trying to write the waterbody_restart_file. I'm not sure why `lastobs_df` is being checked here.

src/model_DAforcing.py
```diff
@@ -369,10 +372,22 @@ def run(self, values: dict):
...
                 lastobs_output_folder
             )
+  
+        # if output_parameters.get('stream_output',{}) is not None and output_parameters.get('stream_output',{}).get('qvd_nudge'):
+        if output_parameters.get('stream_output',{}):
+            
+            t0 = self._t0
+            stream_output_directory =  output_parameters.get('stream_output').get('stream_output_directory')
+            stream_output_type = output_parameters.get('stream_output').get('stream_output_type')
+            stream_output_internal_frequency = output_parameters.get('stream_output').get('stream_output_internal_frequency')
```
shorvath-noaa, line 382: You might need to reformulate these checks. Based on the default values for these parameters, I think a user could cause an error if they have something like
```
stream_output:
    #stream_output_directory:
```
In this case `stream_output_directory` would be `None` which I believe would cause a problem in the write function.

src/model_DAforcing.py
```diff
@@ -369,10 +372,22 @@ def run(self, values: dict):
...
             
-        if output_parameters.get('stream_output',{}).get('qvd_nudge'):
-            #write flowveldepth file
-            pass
+            write_flowveldepth_netcdf(values,
+                                      stream_output_directory,   
+                                      t0,
+                                      1,
```
shorvath-noaa, line 387: This should be defined based on how the BMI engine is being run. We think it'll likely be hourly time intervals, but it doesn't have to be.

src/model_DAforcing.py
```diff
@@ -994,4 +1008,154 @@ def _write_lastobs(
...
+                    ncfile.setncatts(
+                        {
+                            'TITLE': 'OUTPUT FROM T-ROUTE',
+                            'Time step (sec)': f'{stream_output_internal_frequency}',
+                            'model_initialization_time': t0.strftime('%Y-%m-%d_%H:%M:%S'),
+                            'comment': f'The file includes {stream_output_timediff} hour data which includes {len(time_dim)} timesteps',
+                            'code_version': '',
+                        }
+                    )
+                    print(f"{file_name} saved as NetCDF in {stream_output_directory}")
```
shorvath-noaa, line 1160: Can you update this function to be in line with the most recent version in `nhd_io.py`? These print statements, for example, have been put in `LOG.debug()`.

shorvath-noaa (changes requested):

src/model_DAforcing.py
AminTorabi-NOAA: That's true. But should we catch it with pydantic before it starts though?

shorvath-noaa:
@AminTorabi-NOAA is this PR still needed? It looks like it has been superseded by [PR#697](https://github.com/NOAA-OWP/t-route/pull/697)

AminTorabi-NOAA:
@shorvath-noaa No its not needed. Just closed it

PR #716, opened 2023-12-21
parallel computing for storing output

AminTorabi-NOAA:
Update on `write_flowveldepth_netcdf` function for parallel computing. In this setup, each iteration of the loop prepares a job for parallel processing, and the Parallel block at the end executes all these jobs. The helper function  will receive each subset of the data and be responsible for processing it and writing it to the appropriate file format.

## Additions

- helper function added and some part of the `write_flowveldepth_netcdf` function moved there for more readability 

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


src/troute-network/troute/nhd_io.py
```diff
@@ -1948,14 +1948,104 @@ def write_waterbody_netcdf(
...
                               flowveldepth, 
                               nudge, 
                               usgs_positions_id, 
                               t0, 
                               stream_output_timediff, 
                               stream_output_type,
-                              stream_output_internal_frequency = 5):
+                              stream_output_internal_frequency = 5,
+                              cpu_pool = 6, 
+                              use_parallel = True):
```
shorvath-noaa, line 2048: It looks like this `use_parallel` has not been added to output.py when the function gets called? If you want to do it this way, you also need to determine when/how `use_parallel` is determined based on user input. I think you might be able to just drop this though and use only `cpu_pool` in the `if/else` statements. What do you think?

AminTorabi-NOAA: Yes sure. So I can just say if cpu_pul > 1 do parallel otherwise don't. The only caveat is user have to set cpu_pool = 1 in output.py not here (the value 6 here doesn't matter)

shorvath-noaa: cpu_pool is already an input into `nwm_output_generator` in [output.py](https://github.com/shorvath-noaa/t-route/blob/4cd3e498ca3e44941ca93401209ffc791f548b6a/src/troute-nwm/src/nwm_routing/output.py#L49-L65), so it just needs to be added as in input into [nhd_io.write_flowveldepth_netcdf()](https://github.com/shorvath-noaa/t-route/blob/4cd3e498ca3e44941ca93401209ffc791f548b6a/src/troute-nwm/src/nwm_routing/output.py#L205-L212)

shorvath-noaa (approved):

PR #732, opened 2024-01-30
simple fix for empty lastobs_df when streamnudging unknowingly turned on while no usgs data is available

kumdonoaa:
This fix provides a solution when lastobs_df is empty because no usgs data is available while stream nudging is unknowingly activated in the config file. More solid solution will be provided soon.  Also, in diffusive_routing mainstem_list now has a list  of stream segments aligned in order from upstream to downstream.  Moreover, missing topobathy data for a stream segment get filled in by looking up the newly aligned list.

## Additionns
**AbstractRouting.py**
- removed sort() to mainstem_list to align a list of stream segments in order from upstream to downstream
- modified the way to search for an upstream segment id for a given mainstem segment that has no channel xsec topobathy data

**DataAssimilation.py**
- Added a condition not to call new_lastobs function when there is no lastobs related values updated from mc_reach.pyx

**output.py**
- Added a condition not to call lastobs_df_output when there is no lastobs related values updated from mc_reach.pyx

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/src/troute-network/troute/AbstractRouting.py
+++ b/src/troute-network/troute/AbstractRouting.py
@@ -269,7 +269,7 @@ class MCwithDiffusive(AbstractRouting):
         
     def diffusive_domain_by_both_ends_streamid(self, connections, headlink_mainstem, twlink_mainstem, rfc_val, rpu_val):
         # This function build diffusive_domain using given headwater segment IDs at upper and tailwater at lower ends of mainstem.
-        
+
         uslink_mainstem = headlink_mainstem
         dslink_mainstem = 1 # initial value
         mainstem_list =[headlink_mainstem]
@@ -283,8 +283,8 @@ class MCwithDiffusive(AbstractRouting):
                 LOG.debug(f"KeyError: 'connections' does not have a key '{uslink_mainstem}'")
                 return None 
 
-        diffusive_domain = {'links':sorted(mainstem_list), 'rfc': rfc_val, 'rpu': rpu_val, 'upstream_boundary_link_mainstem':[headlink_mainstem]}
-                
+        diffusive_domain = {'links':mainstem_list, 'rfc': rfc_val, 'rpu': rpu_val, 'upstream_boundary_link_mainstem':[headlink_mainstem]}
+   
         return diffusive_domain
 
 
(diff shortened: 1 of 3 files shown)


src/troute-network/troute/DataAssimilation.py
```diff
@@ -238,10 +238,12 @@ def update_after_compute(self, run_results, time_increment):
             - lastobs_df               (DataFrame): Last gage observations data for DA
         '''
         streamflow_da_parameters = self._data_assimilation_parameters.get('streamflow_da', None)
- 
+
         if streamflow_da_parameters:
             if streamflow_da_parameters.get('streamflow_nudging', False):
-                self._last_obs_df = new_lastobs(run_results, time_increment)
+                if not run_results[0][3][0].size == 0:
```
shorvath-noaa, line 244: I think this check needs to be modified. This only checks if the first object in `run_results` has no gage information. But this varies for each object:

```
(Pdb) [rr[3][0].size for rr in run_results]
[2,2,0,2,0,1,0,0,3,0,17,1,1,4,3,1,0,0,0,2,0,1,2,0,0,0,1,0,3,1,2,1,1,0,1,1,0,1,0,1,3,0,0,0,0,0,0,0,3,1,0,11,0,2,0,0,3,1,0,0,0,2,4,0,0,3,0,0,0,0]
(Pdb)
```

src/troute-network/troute/AbstractRouting.py
```diff
@@ -312,10 +312,19 @@ def topobathy_df(self):
...
-                            upstream_mainstem_segment = set(self._diffusive_network_data[tw]['rconn'][mainstem_segment]).intersection(
-                                                                                    self._diffusive_network_data[tw]['mainstem_segs'])
-                            upstream_mainstem_segment = list(upstream_mainstem_segment)[0]
-                            temp_df = self._topobathy_df[self._topobathy_df.index==upstream_mainstem_segment]
+                            temp_df = pd.DataFrame()
+                            position_mainstem_segment =  self._diffusive_network_data[tw]['mainstem_segs'].index(mainstem_segment)
+                            position_upstream_mainstem_segment = position_mainstem_segment
+                            while temp_df.empty:               
+                                try:
+                                    position_upstream_mainstem_segment = position_upstream_mainstem_segment - 1
```
AminTorabi-NOAA, line 320: Why do you decrease it by 1 here? Are the IDs always consecutive?

src/troute-network/troute/DataAssimilation.py
kumdonoaa: what about checking self._last_obs_df? If it is empty in the first place, then we expect nothing will come out for lastobs values from mc_reach.pyx.

src/troute-network/troute/AbstractRouting.py
kumdonoaa: That does not decreasing flowpath ID number but the position of stream segment in the mainstem segment list that has aligned the list from upstream to downstream.

AminTorabi-NOAA: True that's correct. Just a very small suggestion, its more readable to write it like `position_upstream_mainstem_segment -= 1`

src/troute-network/troute/DataAssimilation.py
kumdonoaa: Thanks for your suggestion.

src/troute-network/troute/AbstractRouting.py
kumdonoaa: Thanks and updated.

PR #768, opened 2024-05-23
Update for usace issue

AminTorabi-NOAA:
in the config file `usace_timeslices_folder  : usace_timeslices/` the letter 's' was missing and that was causing the t-rout skip usace_timeslices folder. By adding 's' it would go through the folder but since the usace files are all empty it throw an error and couldn't process it. So the update in `nhd_io.py` checks if the size is zero then return empty dataframe

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/src/troute-network/troute/nhd_io.py
+++ b/src/troute-network/troute/nhd_io.py
@@ -1059,27 +1059,30 @@ def _read_timeslice_file(f):
         stns      = ds.variables['stationId'][:].filled(fill_value = np.nan)
         t         = ds.variables['time'][:].filled(fill_value = np.nan)
         qual      = ds.variables['discharge_quality'][:].filled(fill_value = np.nan)
+   
+    if discharge.size != 0 and stns.size != 0 and t.size != 0:        
+        stationId = np.apply_along_axis(''.join, 1, stns.astype(str))
+        time_str = np.apply_along_axis(''.join, 1, t.astype(str))
+        stationId = np.char.strip(stationId)
         
-    stationId = np.apply_along_axis(''.join, 1, stns.astype(str))
-    time_str = np.apply_along_axis(''.join, 1, t.astype(str))
-    stationId = np.char.strip(stationId)
-    
-    timeslice_observations = (pd.DataFrame({
-                                'stationId' : stationId,
-                                'datetime'  : time_str,
-                                'discharge' : discharge
-                            }).
-                             set_index(['stationId', 'datetime']).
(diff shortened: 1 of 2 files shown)


src/troute-network/troute/nhd_io.py
```diff
@@ -1157,7 +1160,12 @@ def get_obs_from_timeslices(
         for f in timeslice_files:
             jobs.append(delayed(_read_timeslice_file)(f))
         timeslice_dataframes = parallel(jobs)
-        
+
+    all_empty = all(df.empty for tuple in timeslice_dataframes for df in tuple)
+    if all_empty:
+        LOG.debug(f'DataFrames in the list are empty.')
```
shorvath-noaa, line 1166: Can you specify what 'dataframes' or 'list' are being referred to here?

src/troute-network/troute/nhd_io.py
```diff
@@ -1157,7 +1160,12 @@ def get_obs_from_timeslices(
...
+        return pd.DataFrame()
```
shorvath-noaa, line 1167: Does this end the function if activated? Or will the lines in the rest of this function still be run (which would be unnecessary if the dataframes are empty)?

i.e., does this need an `else:` statement here?

AminTorabi-NOAA: it ends the function. if its all empty we don't need to continue, we can just return empty dataframe (similar to else: statement)

JurgenZach-NOAA:
I'm weighing in here, b/c I'm presently working on the related
pydantic check of the gage folders.

We do have situations where all of the USACE files are empty, including our
beloved Lower Colorado test example. I would advise against throwing an
error in such situations. I believe the user needs to be told somewhere if
all the USACE files are empty, but there is no reason not to run the
simulation.

JurgenZach-NOAA:
To be more specific: I understand that we could just turn off reservoir
persistence for usace, but if we have a situation where someone wants
to run the same model over a long and then a shorter time period, and we
happen to not have gage data over the shorter period, then the
simulation would throw an error if we disallow empty gage files. Which
would be difficult to understand for a user.

shorvath-noaa (approved):

PR #779, opened 2024-06-07
fix for diffusive to run with synthetic and topo on NHD

kumdonoaa:
routing_scheme_map in AbtractNetwork.py was updated to run diffusive when using channel topobathy data was turned off in the config yaml file.  Now when "run_hybrid_routing" and "use_natl_xsections" are True and False, respectively, t-route run MC+diffusive both with synthetic xsec data provided by RouteLink file.

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/src/troute-network/troute/AbstractNetwork.py
+++ b/src/troute-network/troute/AbstractNetwork.py
@@ -474,10 +474,10 @@ class AbstractNetwork(ABC):
         routing_type = [run_hybrid, use_topobathy, run_refactored]
 
         _routing_scheme_map = {
-            MCOnly: [False, False, False],
-            MCwithDiffusive: [True, False, False],
-            MCwithDiffusiveNatlXSectionNonRefactored: [True, True, False],
-            MCwithDiffusiveNatlXSectionRefactored: [True, True, True],
+            MCOnly: [[False, False, False]],
+            MCwithDiffusive: [[True, False, False],[True, True, False]],
+            MCwithDiffusiveNatlXSectionNonRefactored: [[True, True, False]],
+            MCwithDiffusiveNatlXSectionRefactored: [[True, True, True]],
             }
         
         # Default to MCOnly routing
@@ -485,9 +485,9 @@ class AbstractNetwork(ABC):
 
         # Check user input to determine the routing scheme
         for key, value in _routing_scheme_map.items():
-            if value==routing_type:
+            if routing_type in value:
                 routing_scheme = key
-        
+
         routing = routing_scheme(self.hybrid_parameters)
 
         (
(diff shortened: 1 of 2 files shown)


src/troute-network/troute/AbstractNetwork.py
```diff
@@ -474,20 +474,20 @@ def initialize_routing_scheme(self,):
         routing_type = [run_hybrid, use_topobathy, run_refactored]
 
         _routing_scheme_map = {
-            MCOnly: [False, False, False],
-            MCwithDiffusive: [True, False, False],
-            MCwithDiffusiveNatlXSectionNonRefactored: [True, True, False],
-            MCwithDiffusiveNatlXSectionRefactored: [True, True, True],
+            MCOnly: [[False, False, False]],
+            MCwithDiffusive: [[True, False, False],[True, True, False]],
+            MCwithDiffusiveNatlXSectionNonRefactored: [[True, True, False]],
```
shorvath-noaa, line 479: I'm confused here, it looks like you added a `[True, True, False]` option (line 478), but that already exists as `MCwithDiffusiveNatlXSectionNonRefactored` (line 479)

shorvath-noaa (changes requested):
It looks like these changes should be in the class `MCwithDiffusiveNatlXSectionNonRefactored` rather than `MCwithDiffusive`. The former is meant to be MC with Diffusive using natural cross-sections, while the latter is MC with Diffusive using synthetic cross-sections.

src/troute-network/troute/AbstractNetwork.py
kumdonoaa: It is true but MCwithDiffusive was mainly written for using topo data while 'use_topobathy'=False.   So, when 'use_topobathy'=False, MCwithDiffusive throws errors in many cases (e.g., L221).   So I guess then the fix can be that we keep _routing_scheme_map as before, but keep the changes I made in AbstractNetwork.py.     Or, we can move L216~L237 to somewhere under class MCwithDiffusiveNatlXSectionNonRefactored?

kumdonoaa: Also, RouteLink file doesn't have column name 'mainstem', we cannot use def _fill_in_missing_topo_data of AbstractRouting.py for NHD.   Furthermore, topobathy data isn't available anymore for NHD.  So, unless communities try to make their own topodata for NHD,  'MConly' and 'MCwithDiffusive (with use_topobathy=False) are only options to choose for NHD.

shorvath-noaa (approved):

PR #780, opened 2024-06-10
`pip>23.1` compliance

aaraney:
[`pip` `23.1`](https://pip.pypa.io/en/stable/news/#id80) changed the default behavior when the `wheel` package is not installed. `troute.network` and `troute.routing` both require `wheel` at build time and previously did not contain `pyproject.toml` files.

> When the `wheel` package is not installed, `pip` now uses the default build backend instead of `setup.py install` and `setup.py develop` for project without `pyproject.toml`. ([#8559](https://github.com/pypa/pip/issues/8559))

The PR addresses these issues and makes it possible to install `troute` using the latest `pip` version (`24.0`).

Fixes #621.

PR #782, opened 2024-06-12
Add build dep `wheel` to `requirements.txt`

aaraney:
Should have added this in #780. This should fix [the CI docker image build failures](https://github.com/NOAA-OWP/t-route/actions/runs/9457679519/job/26051934655).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+wheel
 numpy
 Cython<3.0.4
 pandas >=1.1.5, <=2.2.0


aaraney:
Okay, it looks like when I manually installed `wheel` in the docker image before building `t-route`, that got rid of the build issues (see [action](https://github.com/NOAA-OWP/t-route/actions/runs/9486367062/job/26140536678?pr=782)).

Since we will pull in `wheel` from `requirements.txt` after this is merged, I am going to remove the "manual" install of `wheel` in the dockerfile. As a result, the action will fail again when run for the PR, but once we merge the issue should go away.

aaraney:
Okay, just removed the temporary commit. We should be good to merge now.

kumdonoaa (approved):

PR #816, opened 2024-07-31
speeding up Parquet_output function by vectorizing

AminTorabi-NOAA:
Reworking a `_parquet_output_format_converter` function to speeding it up by vectorizing and using numpy array. Previously it was taking around **50 sec**. Can see in below picture.
image

But by vectorizing it. It improved to **1 second**.
image

The result is also checked its the same.
Original function output:
image

Updated function output:
image

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/src/troute-nwm/src/nwm_routing/output.py
+++ b/src/troute-nwm/src/nwm_routing/output.py
@@ -83,7 +83,7 @@ def _parquet_output_format_converter(df, start_datetime, dt, configuration, pref
     variables = np.array([t[1] for t in times])
 
     # Preallocate arrays
-    location_ids_repeated = np.repeat(location_ids, num_time_variables)
+    location_ids_repeated = np.tile(location_ids, num_time_variables)
     value_time = np.empty(num_records, dtype='datetime64[us]')
     variable_names = np.empty(num_records, dtype=object)
     units = np.empty(num_records, dtype=object)
@@ -110,7 +110,7 @@ def _parquet_output_format_converter(df, start_datetime, dt, configuration, pref
         'reference_time': start_datetime.date(),
         'configuration': configuration
     })
-
+    import pdb;pdb.set_trace()
     return timeseries_df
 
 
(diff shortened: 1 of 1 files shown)


src/troute-nwm/src/nwm_routing/output.py
```diff
@@ -64,28 +64,53 @@ def _parquet_output_format_converter(df, start_datetime, dt, configuration, pref
...
-    timeseries_df['location_id'] = timeseries_df['location_id'].astype('string')
-    timeseries_df['location_id'] = prefix_ids + '-' + timeseries_df['location_id']
-    timeseries_df['value'] = timeseries_df['value'].astype('double')
-    timeseries_df['reference_time'] = timeseries_df['reference_time'].astype('datetime64[us]')
-    timeseries_df['value_time'] = timeseries_df['value_time'].astype('datetime64[us]')
 
+    # Prepare the location_id with prefix
+    df.index.name = 'location_id'
+    df.reset_index(inplace=True)
+    location_ids = prefix_ids + '-' + df['location_id'].astype(str)
```
shorvath-noaa, line 73: If `prefix_id` is `'nex'`, I assume the user would want nexus point values rather than flowpath values. In that case the `flowveldepth` data frame (which is indexed by flowpath ID) should be aggregated by downstream nexus point ID (similar to what happens in `stream_output`). Can you add that here?

shorvath-noaa (changes requested):
Overall this looks good, I'm glad this could be sped up. See my one comment about aggregating values by nexus point ID.

src/troute-nwm/src/nwm_routing/output.py
AminTorabi-NOAA: Sure, we can do that need to pass the RoutLink.nc and make a crosswalk between 'from' and 'to'. 
But, I think that would be a separate PR. This PR is just for speeding up an existing function.

shorvath-noaa (approved):

PR #821, opened 2024-08-07
update default values for parquet output

shorvath-noaa:
Add default values for `prefix_ids` and `configurations`.

## Additions

-

## Removals

-

## Changes
**output_parameters.py**
- `configuration: Optional[str] = 'None'`
- `prefix_ids: Optional[str] = 'wb-'`

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ ] PR has an informative and human-readable title
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code can be automatically merged (no conflicts)
- [ ] Code follows project standards (link if applicable)
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [ ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Target Environment support

- [ ] Windows
- [ ] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS
- [ ] Flexible from small to large screens
- [ ] No linting errors or warnings
- [ ] JavaScript tests are passing

PR #825, opened 2024-08-12
fix: parquet `prefix_ids` defaults to `wb`

aaraney:
Remove unnecessary `-` suffix from `prefix_ids` (I should have caught this when reviewing #821).

#821 changed the default `prefix_ids` to `wb-`. The parquet output handling code sets the feature location to `{prefix_ids}-{location_id}`. So, the default would now be `wb--{location_id}`. We want `wb-{location_id}`. This fixes that.

related to #821


shorvath-noaa (approved):

aaraney:
Thanks, @shorvath-noaa!

Issue #604, opened 2023-02-22
drop down to sqlite3 for gpkg read to avoid gdal dependencies


jameshalgren:
https://github.com/NOAA-OWP/t-route/blob/01c713acb02e03e1d4222b62c398c30981485030/src/troute-network/troute/HYFeaturesNetwork.py#L16-L20

In most cases (in every case?) for t-route, we are discarding the geometry data and only use the geo-enabled packages for convenience in reading the files where the network connectivity information is stored. 

To avoid unnecessary (see note) dependencies, we could update this function to look something like the following: 
``` py
## TESTED; See note below
import sqlite3
import pandas as pd

def read_geopkg(file_path):
    flowpaths = read_geopkg_layer_sqlite3(file_path, layer="flowpaths")
    attributes = read_geopkg_layer_sqlite3(file_path, layer="flowpath_attributes")
    #merge all relevant data into a single dataframe
    return pd.merge(flowpaths, attributes, on='id')

def read_geopkg_layer_sqlite3(file_path, layer):
    con = sqlite3.connect(file_path)
    return pd.read_sql_query("SELECT * from " + layer, con)
    
```

_Note_: More accurately, "troublesome" dependencies. Gdal has been causing some trouble on the arm64 builds and if we can circumvent those issues, we gain (potentially) quite a bit of performance. 
_Note2_: The geopandas read is (I believe) lazy, so this direct sql read with the connection closing will carry a performance hit, potentially, especially for very large flowpath tables.

jameshalgren:
Tested this by using the old function and the new function to read a geopackage for the 03w VPU:
```
wget -P 03w -c https://nextgen-hydrofabric.s3.amazonaws.com/v1.2/nextgen_03W.gpkg
```

Then compare the outputs:
```py
d = read_geopkg("nextgen_03W.gpkg")
d2 = read_geopkg_sqlite3("nextgen_03W.gpkg")
d.drop(["geometry"],axis=1) == d2.drop(["geom", "fid_x", "fid_y"],axis=1)
```

The only difference is the geometry.

jameshalgren:
ping @ZacharyWills

jameshalgren:
> Your sql statement is a special case that has its own convenience function, `pd.read_sql_table`.
> 
> ```python
> pd.read_sql_table('flowpaths', con, index_col='id')
> ```

Python 3.10+

PR #833, opened 2024-08-22
remove geopandas and fiona + bugfixes

JoshCu:
# Geopandas
geopandas (more specifically it's fiona dependency) causes issues when wheels aren't available and it attempts to build from source on arm64/aarch64. building fiona is currently preventing NGIAB from building and has caused issues with arm NGIAB in the past.
As far as I can tell it's just used to read non-geometric data from geopackages and non-geometric data from geojson files.

```python
#this
gdf = gpd.open_file("conus.gpkg", layer="divides")
# can be done like this
with sqlite3.connect("conus.gpkg") as conn:
    df = pd.read_sql_query("SELECT * FROM divides", conn)
``` 
The geometry is an unpleasant blob that would need to be decoded [something like this](https://github.com/CIROH-UA/NGIAB_data_preprocess/blob/9809b532629e2ce6cd93b37ce0c6790dd6dea102/modules/data_processing/gpkg_utils.py#L70), but it doesn't appear to be used anywhere so we could probably not bother select it from the geopackage in the first place to save some time and memory

Similarly, geojson is read into a normal pandas df, the geometry is less mangled than the geopackage but still not a proper geometry object. The geometry doesn't seem to be used anywhere, so I just left some comments explaining the format if anyone in the future does need to use it.

I haven't done performance tests on this, but it should also be faster than geopandas. 

### Some similar work
https://github.com/NOAA-OWP/t-route/issues/604
https://github.com/NOAA-OWP/t-route/pull/568

## Removals

- All imports and use of geopandas and fiona

## Changes

- Geopackages now open with pandas + sqlite3 and are **normal dataframes**
- Geojson now opens by reading the raw json and reformatting the contents into a normal pandas dataframe

## Testing & Screenshots

It worked on a small simulation generated with ngiab_data_preprocess, running inside NGIAB
image

# Additional Bugfixes
In order to test this using geopackages generated by ngiab_data_preprocess (v20.1 of the hydrofabric) I had to modify the regex and pandas table merging. 

## Regex
The original regex would match both flowpaths and flowpath_attributes as flowpaths.
image
`matched_layers: {'flowpaths': 'flowpath_attributes', 'flowpath_attributes': 'flowpath_attributes', 'lakes': None, 'nexus': 'nexus', 'network': None}`
adding $ to check for the end of line fixed this.

## table merging
I noticed this while debugging the regex issue; trying to merge two dataframes when either one of them didn't have the "id" column was causing errors. Now it checks to see what combination of the flowpaths and flowpath_attributes exist and merge accordingly.

### DIFF
--- a/src/troute-network/troute/HYFeaturesNetwork.py
+++ b/src/troute-network/troute/HYFeaturesNetwork.py
@@ -28,7 +28,7 @@ def find_layer_name(layers, pattern):
         if re.search(pattern, layer, re.IGNORECASE):
             return layer
     return None
-  
+
 def read_geopkg(file_path, compute_parameters, waterbody_parameters, cpu_pool):
     # Retrieve available layers from the GeoPackage 
     with sqlite3.connect(file_path) as conn:
@@ -76,7 +76,7 @@ def read_geopkg(file_path, compute_parameters, waterbody_parameters, cpu_pool):
         try:
             with sqlite3.connect(file_path) as conn:
                 has_spatial_metadata = False
-                # try and get the name of the geometry column and it's crs
+                # try and get the name of the geometry column and its crs
                 geometry_columns = conn.execute(f"""
                 SELECT c.column_name,g.definition
                 FROM gpkg_geometry_columns AS c
@@ -89,7 +89,9 @@ def read_geopkg(file_path, compute_parameters, waterbody_parameters, cpu_pool):
                     crs = geometry_columns[0][1]
 
                 if has_spatial_metadata:
-                    # select everything from the layer, + the midpoint of it's bounding box
+                    # select everything from the layer, + the midpoint of its bounding box
+                    # decoding the geometry blob can be done as it's just gpkg header + WKB
+                    # the rtree table calculation is much faster
                     sql_query = f"""SELECT d.*,
                         (r.minx + r.maxx) / 2.0 AS lon,
                         (r.miny + r.maxy) / 2.0 AS lat
@@ -170,8 +172,7 @@ def read_geojson(file_path):
     df = pd.json_normalize(data,max_level=1)
     df.columns = df.columns.str.replace('properties.','')
     df = df.drop(columns=['type'])
-    # Geometry seems to be unused or dropped, in case it is needed:
-    # geometry type e.g. MULTIPOLYGON, is stored in geometry.type
+    # Geometry type e.g. MULTIPOLYGON, is stored in geometry.type
     # and the coordinates are stored in geometry.coordinates
     # crs stored in data['crs'] e.g.
     # data['crs'] = { "type": "name", "properties": { "name": "urn:ogc:def:crs:EPSG::5070" } }
--- a/src/troute-network/troute/HYFeaturesNetwork.py
+++ b/src/troute-network/troute/HYFeaturesNetwork.py
@@ -69,19 +69,27 @@ def read_geopkg(file_path, compute_parameters, waterbody_parameters, cpu_pool):
     if hybrid_parameters.get('run_hybrid_routing', False) and 'nexus' not in layers_to_read:
         layers_to_read.append('nexus')
 
-    # Function that read a layer from the geopackage
+    # Function that reads a layer from the geopackage
     def read_layer(layer_name):
         if not layer_name:
             return pd.DataFrame()
         try:
             with sqlite3.connect(file_path) as conn:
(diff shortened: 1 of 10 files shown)


shorvath-noaa:
The geometries are actually used in a couple places (that's why the GitHub Actions tests are failing). t-route uses the geometries to retrieve lat/lon information for [nexus points](https://github.com/NOAA-OWP/t-route/blob/9d9d711435c553633bedc6056c9827dfb9e7843e/src/troute-network/troute/AbstractNetwork.py#L921-L935) (which are intended to be passed to the model engine via BMI functions at some point to then pass along to coastal models) and [waterbodies](https://github.com/NOAA-OWP/t-route/blob/9d9d711435c553633bedc6056c9827dfb9e7843e/src/troute-network/troute/HYFeaturesNetwork.py#L530-L546) (to mimic the LAKEOUT files that WRF-Hydro creates, which include waterbody lat/lon). There might be additional places where these packages are used, but I'd need to take a closer look.

So, if we want to remove geopandas and fiona, we would need to either:
1) Find another geospatial manipulation package to do these operations.
2) Ask if lat/lon info can be included in the hydrofabric somewhere other than the `geometry` columns.

src/troute-network/troute/HYFeaturesNetwork.py
```diff
@@ -29,14 +28,19 @@ def find_layer_name(layers, pattern):
         if re.search(pattern, layer, re.IGNORECASE):
             return layer
     return None
-
+
```
jameshalgren, line 31: minor whitespace diff here...

src/troute-network/troute/HYFeaturesNetwork.py
```diff
@@ -65,15 +69,41 @@ def read_geopkg(file_path, compute_parameters, waterbody_parameters, cpu_pool):
...
+                JOIN gpkg_spatial_ref_sys AS g ON c.srs_id = g.srs_id 
+                WHERE c.table_name = '{layer_name}'""").fetchall()
+                
+                if len(geometry_columns) > 0:
+                    has_spatial_metadata = True
+                    geom_column = geometry_columns[0][0]
+                    crs = geometry_columns[0][1]
+
+                if has_spatial_metadata:
+                    # select everything from the layer, + the midpoint of it's bounding box
```
jameshalgren, line 92: the midpoint of *its* bounding box.

src/troute-network/troute/HYFeaturesNetwork.py
JoshCu: oops, I must not have had black installed

AminTorabi-NOAA:
I test the performance, there is a good improvement using sql

AminTorabi-NOAA (approved):

JoshCu:
I didn't want to litter the code with too many comments, so I'll leave some notes here:

- The source I used for geospatial index rtree info is this http://www.geopackage.org/spec120/#extension_rtree
- For blob decoding http://www.geopackage.org/spec120/#gpb_format, I opted not to use it here because it's slower and requires shapely for wkb.loads to and pyproj to do the transformation
- It's possible to read the geometry blob header and get the bounding box without having to decode the wkb https://github.com/CIROH-UA/NGIAB_data_preprocess/blob/35ad58c5f776593c2ca92711555815955e16896a/modules/data_processing/gpkg_utils.py#L93
  - This would be needed if the rtree tables aren't there (I don't think they're required to be)
  - If the rtree tables are missing and need adding, it's not difficult, but it is messy. It requires sqlite3 to be built with rtree enabled (there by default on the sqlite3 versions I've tried), creating the rtree tables, populating them, adding triggers to maintain them, and manually adding some other metadata to gpkg_ogr_contents, gpkg_contents, and gpkg_geometry_columns

Geopackages property fields in the geopackages aren't always indexed: This doesn't seem to be relevant here, but if you wanted to select a subset of the entries by id:
```SQL
SELECT * FROM divides WHERE id=wb-9999
```
Takes ~1s on the conus gpkg. If there's an index on id, it takes <30ms
```SQL
CREATE INDEX "diid" ON "divides" ( "id" ASC )
```
I overlooked this on my gpkg and lost an afternoon to performance testing different access types

JoshCu:
This change was quite large and moved away from a well maintained package to hard coded operations. My main goal was to fix t-route for NGIAB and that can be achieved with two less intrusive changes #838 and #839.

PR #837, opened 2024-09-03
RTX Code Release (PI-3): Enhancements to T-Route so it can be used for OWP's Replace and Route

taddyb33:
The FIM-S team at RTX has been making additions to T-Route locally to support the Replace and Route tool (https://github.com/NOAA-OWP/hydrovis/tree/ti/Source/RnR). We want to push these changes back to the base t-route repo as per contractual requirements and for the hydrological community to benefit as a whole. 

We developed a FastAPI endpoint for T-Route to be run as a service through `localhost:8004` using parameters and forcings. We also have developed a procedure to push a built T-Route image to a GitHub container registry to be used by other services.

## Additions

- exposure of the `main_v04` function within T-Route
- A FastAPI endpoint (`/api/v1/flow_routing/v4`) to utilize inputs and shared volume mounts to run T-Route as a Service
- an API doc spec and Quickstart file to get the container up and running
- A test compose file, `test_compose.yaml` and test data `test/api/data`
- A docker compose setup for T-Route using the `Dockerfile.troute_api` and `compose.yaml`

## Removals

-

## Changes

- exposure of the `main_v04` function within following file: `src/troute-nwm/src/nwm_routing/__init__.py`
```python
from .__main__ import main_v04
```

## Testing

1. Follow the steps within `test/api/README.md`

## Notes

- For an example docker build, see https://github.com/taddyb33/t-route-dev/pkgs/container/t-route-dev

## Todos

-

## Checklist

- [ ] Windows
- [x] Linux
- [ ] Browser

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


test/api/README.md
```diff
@@ -0,0 +1,21 @@
+# Testing the T-Route FastAPI Extension
+
+The following folder contains data files that are to be used to test the T-Route FastAPI code within src/app
+
+To use these files, follow the steps below:
+
+1. Copy the `test_compose.yaml` file in the root project dir
```
AminTorabi-NOAA, line 7: Can you be more specific with root dir? Do you mean in the `src` folder or lets say we are running LowerColorado in that folder

test/api/README.md
```diff
@@ -0,0 +1,21 @@
...
+4. Enter the following parameters into the `/api/v1/flow_routing/v4` endpoint
+- lid=CAGM7
+- feature_id=2930769
+- hy_id=1074884
+- initial_start=0
+- start_time=2024-08-24T00:00:00
+- num_forecast_days=5
+5. Click execute
+6. A Status 200 code means the run ran, and test/api/data/troute_output will be populated in the `{lid}/` folder
+
```
AminTorabi-NOAA, line 19: I am getting below error
image

test/api/README.md
taddyb: Sure, I changed this to the following line:
```
Copy the `test_compose.yaml` file in the base project dir (/t-route)
```

Let me know if I need to be more specific

taddyb33: Sure, I changed this to the following line:
```
Copy the `test_compose.yaml` file in the base project dir (/t-route)
```

Let me know if this should be more verbose.

_Edit: originally commented with my personal GitHub account. had to switch to my RTX one_

test/api/README.md
taddyb33: Can you post a screenshot of the the terminal that has docker compose running (where you ran `docker compose -f test_compose.yaml up`)?

I just ran through the test case and got the following output, which matches what T-Route would show if I were to be running the code locally:

My GET request looks like:
```shell
troute-1  | INFO:     172.22.0.1:54084 - "GET /api/v1/flow_routing/v4/?lid=CAGM7&feature_id=2930769&hy_id=1074884&initial_start=0&start_time=2024-08-24T00%3A00%3A00&num_forecast_days=5 HTTP/1.1" 200 OK
```
image

test/api/README.md
AminTorabi-NOAA: The issue was that the `test_compose.yaml` file was in the wrong directory, but it's working now.

A couple more questions/comments:

When I run it, a browser opens to `http://localhost:8000/`, but I have to manually change it to `http://localhost:8004/docs` to display the correct page. Is there a way to open the correct page directly?

Could you provide an example of how to run the Lower Colorado test using `//t-route/test/LowerColorado_TX_v4/test_AnA_V4_HYFeature.yaml`?

test/api/README.md
AminTorabi-NOAA: The issue was that the `test_compose.yaml` file was in the wrong directory, but it's working now.

A couple more questions/comments:

When I run it, a browser opens to `http://localhost:8000/`, but I have to manually change it to `http://localhost:8004/docs` to display the correct page. Is there a way to open the correct page directly?

Could you provide an example of how to run the Lower Colorado test using `//t-route/test/LowerColorado_TX_v4/test_AnA_V4_HYFeature.yaml`?

Note: Previously typed in wrong place

taddyb33: > When I run it, a browser opens to http://localhost:8000/, but I have to manually change it to http://localhost:8004/docs to display the correct page. Is there a way to open the correct page directly?

I changed the `test_compose.yaml` file to point to port 8000. The reason it was on port 8004, rather than 8000 to begin with, was because of testing with the Replace and Route application. Since this compose file is local to T-Route, changing the FastAPI port back to 8000 will have no effect on RnR.

kumdonoaa: As Amin mentioned, testing your application with //t-route/test/LowerColorado_TX_v4/test_AnA_V4_HYFeature.yaml would be more relevant for us to review your PR, as it will help determine whether the T-Route Docker image produces the same output as the current T-Route version and behaves as expected in terms of compute speed and stability.

taddyb33: Thanks, @kumdonoaa and @AminTorabi-NOAA for the suggestion. I just pushed updates to the `test/api/README.md` file for how to run this API with the `LowerColorado` data. 

Some things to note:
1. I copied the `LowerColorado` channel forcings and v20.1 hydrofabric gpkg to the same directories that the `test/api/data` is located. This was to prevent having to change the mounted volume location within the `test_compose.yaml` file
2. You'll notice in the instructions I added (see below) the `feature_id` and `hy_id` are the same. I put this this way because of how RnR calls this API. RnR saves it's geopackages based on their `feature_id of their downstream point`, and thus we pass in the feature ID to t-route so it knows where to look for the gpkg data. In the `LowerColorado` case, we just want to make sure we know where the data. 
```md
- lid=LowerColorado
- feature_id=2420801
- hy_id=2420801
- initial_start=0
- start_time=2023-04-01T00:00:00
- num_forecast_days=2
```
3. The num_forecast days is set to 2 since the data is from 4/01/23-4/03/23

taddyb33:
FYI: Thanks @hellkite500 for pointing this out, but the additions count, and files added, for this PR is overbloated due to copying the Lower Colorado data into `test/api/data`. 

I'm looking into other methods for reading this data that doesn't result in unneccessary file duplication. If I can't find any, I'll make a custom `compose.yaml` file for the LowerColorado data

See below for the additions count from a few LowerColorado test files

image

AminTorabi-NOAA:
@taddyb33 Thanks for the update on LowerColorado. I think the issue lies with the folder structure. We don’t want to copy forcing, USGS, or USACE files into the new folder, and we prefer not to include them in PRs unless absolutely necessary. This increases the clone size and complicates collaboration for other teams.

- Is it possible for your PR to be flexible with our config file (for example `/test_AnA_V4_HYFeature.yaml`), so we can specify any folder path for forcing files, USGS, etc., or have it read directly from our config? I noticed you connected `rfc_channel_forcings` and `rfc_geopackage_data` and copied the gpkg and forcings into those folders, but we’d rather provide the file locations ourselves.

- I saw the pop-up window allows only a few variables, like `start_time`, to be changed, whereas the config files contain many more. Should we be changing those in the `test_AnA_V4_HYFeature.yaml` file, and only adjust `start_time` in the pop-up?

taddyb33:
I sync'd this branch with the latest T-Route commit, and the duplicated test data was removed, bringing the total additions back to what it should be. Answering the above questions:

> Is it possible for your PR to be flexible with our config file 

As of right now the API does not have integration for static configuration files. Since this couples with Replace and Route, I don't want to accidentally break something from that side of things. 

> I noticed you connected rfc_channel_forcings and rfc_geopackage_data and copied the gpkg and forcings into those folders

That data format is related to how I coded replace and route's file directory. `rfc_channel_forcings` is where the formatted NWPS data is, `rfc_geopackage_data` is the v20.1 gpkg folders

> I saw the pop-up window allows only a few variables, like start_time, to be changed,

The way that the API works is we use a `base_config.yaml` file and modify it using inputs from the POST request. We would need to add more logic to make the API work better with how a user would call it

src/app/api/services/initialization.py
```diff
@@ -0,0 +1,167 @@
...
+
+import geopandas as gpd
+import numpy as np
+import pandas as pd
+import yaml
+
+from app.core.settings import Settings
+
+
+def edit_yaml(original_file: Path, params: Dict[str, str], restart_file: Path) -> Path:
```
hellkite500, line 15: This function is very much specific to the R&R t-route usage semantics, and also seems to miss an opporutnity to use the `troute.config` data models to handle the manipulation/updates, which would alleviate the need for this function to link a yaml dict with a params dict.

src/app/api/services/initialization.py
```diff
@@ -0,0 +1,167 @@
...
+
+    Returns:
+    --------
+    Dict[str, str]:
+        The parameters to be added to the t-route config file
+    """
+    dt = datetime.strptime(start_time, "%Y-%m-%dT%H:%M:%S")
+    start_datetime = dt.strftime("%Y-%m-%d_%H:%M")
+
+    nts = 288 * num_forecast_days
```
hellkite500, line 106: It wouldn't be too difficult to refactor this to use forecast hours as an input, would make this more generally useful for a wider range of potential applications

src/app/api/services/initialization.py
```diff
@@ -0,0 +1,167 @@
...
+    --------
+    Path:
+        The path to the t-route restart file
+    """
+    start_datetime = datetime.strptime(params["start_datetime"], "%Y-%m-%d_%H:%M")
+    formatted_datetime = start_datetime.strftime("%Y-%m-%d_%H:%M")
+
+    gdf = gpd.read_file(params["geo_file_path"], layer="network")
+    mask = gdf["divide_id"].isna()
+    keys = [int(val[4:]) for val in set(gdf[~mask]["divide_id"].values.tolist())]
```
hellkite500, line 141: I strongly encourage using string split semantics instead of assuming prefix lengths are always going to be the same.

src/app/api/services/initialization.py
```diff
@@ -0,0 +1,167 @@
...
+            "qu0": discharge_upstream,
+            "qd0": discharge_downstream,
+            "h0": height,
+        }
+    )
+    df.set_index("key", inplace=True)
+    restart_path = Path(settings.restart_path.format(params["lid"]))
+    restart_path.mkdir(exist_ok=True)
+    restart_full_path = restart_path / settings.restart_file.format(formatted_datetime)
+    df.to_pickle(restart_full_path)
```
hellkite500, line 166: Why pickle the dataframe?  This is probably one of the least portable serialization mechanics that pandas supports.  I supposed this is like due to how the restart is deserialized by some other part of t-route that would need to be changed, but if we are to a point where we want to consider supporting external applications providing restart files, then pickle shouldn't be the format since it can only be created and ready by python (with some additional caveats about version comaptibiltiy)

src/app/core/__init__.py
```diff
@@ -0,0 +1,15 @@
+from functools import lru_cache
+
+from app.core.settings import Settings
+
+
+@lru_cache
+def get_settings() -> Settings:
+    """Instantiating the Settings object using LRU caching
```
hellkite500, line 8: What does the LRU provide here, and what constitutes a cache hit and a cache miss?

src/app/core/base_config.yaml
```diff
@@ -0,0 +1,98 @@
...
+            mainstem: 'mainstem'
+    waterbody_parameters:
+        #----------
+        break_network_at_waterbodies: True  
+#--------------------------------------------------------------------------------
+compute_parameters:
+    #----------
+    parallel_compute_method: by-subnetwork-jit-clustered # serial
+    compute_kernel         : V02-structured
+    assume_short_ts        : True
```
hellkite500, line 39: It could be helpful to at least point out how the API could be expanded to provide mechanics to expose other configuration variables to be manipulated by the API, which may help the maintainers unfamiliar with fastapi see what long term maintenance and enahncements may look like

src/app/core/base_config.yaml
```diff
@@ -0,0 +1,98 @@
...
+#     test_output : output/lcr_flowveldepth.pkl
+#     lite_restart:
+#         #----------
+#         lite_restart_output_directory: restart/
+#     lakeout_output: lakeout/
+    stream_output : 
+        stream_output_directory: /t-route/output/{}
+        stream_output_time: 1 #[hr]
+        stream_output_type: '.nc' #please select only between netcdf '.nc' or '.csv' or '.pkl'
+        stream_output_internal_frequency: 60 #[min] it should be order of 5 minutes. For instance if you want to output every hour put 60
```
hellkite500, line 98: More cowbell!

src/app/schemas.py
```diff
@@ -0,0 +1,22 @@
+"""Author: Tadd Bindas"""
+
+from pydantic import BaseModel
+
+
+class TRouteOuput(BaseModel):
```
hellkite500, line 6: TRouteOutput is perhaps a misleading name implying this class/return contains actually routed output data.  I would rename this to something like `TRouteStatus` since this is really just status messaging about the run, but doesn't contain model output.

src/troute-nwm/src/nwm_routing/__init__.py
```diff
@@ -0,0 +1 @@
+from .__main__ import main_v04
```
hellkite500, line 1: More Cowbell!

test/api/test_compose.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - type: bind
+        source: ./src/app/core
+        target: /t-route/src/app/core
+    command: sh -c ". /t-route/.venv/bin/activate && uvicorn app.main:app --host 0.0.0.0 --port 8000"
+    healthcheck:
+      test: curl --fail -I http://localhost:8000/health || exit 1
+      interval: 30s
+      timeout: 5s
+      retries: 3
+      start_period: 5s
```
hellkite500, line 24: More cowbell!

test/api/README.md
```diff
@@ -0,0 +1,22 @@
+# Testing the T-Route FastAPI Extension
```
hellkite500, line 1: Would like to ultimately see some automated unit tests of the API -- setup the instance with the test_compose inside a test framework's init/setup (e.g. pytest) and and create then send requests to the server and verify the expected return messages at least.

compose.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - type: bind
+        source: ./src/app/core
+        target: /t-route/src/app/core
+    command: sh -c ". /t-route/.venv/bin/activate && uvicorn app.main:app --host 0.0.0.0 --port 8000"
+    healthcheck:
+      test: curl --fail -I http://localhost:8000/health || exit 1
+      interval: 30s
+      timeout: 5s
+      retries: 3
+      start_period: 5s
```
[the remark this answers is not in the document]
taddyb: This is a great addition

taddyb:
Thanks @hellkite500 for the review. I'm finally getting back around to completing these tasks and will post updates upon completion with additional context in the PR. Planning on breaking the commits into:
- Generic case with staged data
- General comments (cowbell)
- Documentation

taddyb:
Also, for context. Adding a 👍 reaction to a PR comment means that I've addressed it. Trying to make sure I hit all of the comments and cowbells

Dockerfile.troute_api
```diff
@@ -0,0 +1,26 @@
+FROM rockylinux:9.2 as rocky-base
+RUN yum install -y epel-release 
+RUN yum install -y netcdf netcdf-fortran netcdf-fortran-devel netcdf-openmpi
+
+RUN yum install -y git cmake python python-devel pip 
+
+WORKDIR "/t-route/"
+COPY . "/t-route/"
+
+RUN ln -s /usr/lib64/gfortran/modules/netcdf.mod /usr/include/openmpi-x86_64/netcdf.mod
```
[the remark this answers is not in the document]
taddyb: just an FYI, this was taken from `docker/Dockerfile.troute` https://github.com/NOAA-OWP/t-route/blob/master/docker/Dockerfile.troute#L10C1-L10C88

if I'm making a change to the troute_api container, then the regular container should reflect the same addition

src/app/core/__init__.py
taddyb: This was carried over from a previous template. Removing.

build.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+if ! docker login --username ${GH_USERNAME} --password ${GH_TOKEN} ghcr.io; then
```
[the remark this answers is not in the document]
taddyb: This script is used to push a built T-Route container to their Github Container Registry. After more thought, I think it's best to delete this script and use CI/CD with github secrets set by maintainers to update docker containers rather than manual builds using ENV credentials.

thoughts?

data/troute_restart/.gitkeep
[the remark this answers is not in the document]
taddyb: This was for having a place within the container for restart files to be saved. I'm going to look into generating this inside the container using the dockerfile

taddyb:
I'm pretty much finished up with addressing the comments (thanks @hellkite500 for the review). I'll be uploading another commit once my tests for the generic LowerColorado testing pass (where you can point to a test file and run T-Route through the API) and tagging reviewers once that's there. 

While there were some great comments above, I don't think all of them will be able to make this PR as it means I have to then add scope to Replace and Route. Those additions include:
> This function is very much specific to the R&R t-route usage semantics, and also seems to miss an opporutnity to use the troute.config data models to handle the manipulation/updates, which would alleviate the need for this function to link a yaml dict with a params dict.

> It wouldn't be too difficult to refactor this to use forecast hours as an input, would make this more generally useful for a wider range of potential applications

> It could be helpful to at least point out how the API could be expanded to provide mechanics to expose other configuration variables to be manipulated by the API, which may help the maintainers unfamiliar with fastapi see what long term maintenance and enahncements may look like

We have work scoped out for more T-Route testing support and enhancements to replace and route. I would like to address these in future commits and version control the final working product rather than keep an ever growing PR.

kumdonoaa:
It didn't pass Github Actions.

taddyb:
> It didn't pass Github Actions.

From what I can see, it needs an approval from a maintainer to run actions
- https://github.com/NOAA-OWP/t-route/actions/runs/10779259029 (The first PR)
- https://github.com/NOAA-OWP/t-route/actions/runs/11256909963 (The latest code commit)

taddyb:
@kumdonoaa Based on our CIROH Science Meeting conversations, I added a general dockerfile to support the T-Route API. It's located in `docker/Dockerfile.troute_api`

I also moved all compose.yaml files into `docker/` and added different .env files to manage the mounted volumes and ports. 

This should make testing, and running the API, easier. 

Lastly, docs have been updated in `docker/api/api_docs.md` to reflect these changes

kumdonoaa:
> @kumdonoaa Based on our CIROH Science Meeting conversations, I added a general dockerfile to support the T-Route API. It's located in `docker/Dockerfile.troute_api`
> 
> I also moved all compose.yaml files into `docker/` and added different .env files to manage the mounted volumes and ports.
> 
> This should make testing, and running the API, easier.
> 
> Lastly, docs have been updated in `docker/api/api_docs.md` to reflect these changes

kumdonoaa:
@taddyb It seems this file doesn't exist any more at here: test/api/README.md.   Also no folder of docker/api/.

taddyb:
> @taddyb It seems this file doesn't exist any more at here: test/api/README.md. Also no folder of docker/api/.

Apologies, I mistyped the path name. 

1. Docs are available at `doc/api/api_docs.md`, _not_ `docker/api/api_docs.md`

2. I re-added the `README.md` file in the `test/api/` folder with descriptions on how to test the endpoints.

taddyb:
Closing this Pull Request as newer code has been introduced in #850 that contains these changes, with v2.2 Hydrofabric integration.

PR #838, opened 2024-09-03
Geopackage read bugfix

JoshCu:
## Fixes regex matching both flowpaths and flowpath_attributes to flowpath
( specifically geopackages with both flowpaths and flowpath_attributes tables )
This adds a check for an end of sequence character after flowpath(s) so that flowpaths_attributes doesn't get mapped to flowpaths

Additionally adds logic to prevent the error that would occur from attempting to merge an empty dataframe

separated out from #833 as this is a bugfix and that is a larger fundamental change.

### DIFF
--- a/src/troute-network/troute/HYFeaturesNetwork.py
+++ b/src/troute-network/troute/HYFeaturesNetwork.py
@@ -93,13 +93,18 @@ def read_geopkg(file_path, compute_parameters, waterbody_parameters, cpu_pool):
     if 'link' in flowpath_attributes_df.columns:
         flowpath_attributes_df.rename(columns={'link': 'id'}, inplace=True) 
      
-    # Merge flowpaths and flowpath_attributes 
-    flowpaths = pd.merge(
-        flowpaths_df, 
-        flowpath_attributes_df, 
-        on='id', 
-        how='inner'
-    )
+    # Merge flowpaths and flowpath_attributes
+    if not flowpath_attributes_df.empty and not flowpaths_df.empty:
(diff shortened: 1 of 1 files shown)


AminTorabi-NOAA:
Can you also make a PR for #833. I think it was great idea using sqlite, it's much faster using that

jameshalgren:
> Can you also make a PR for #833. I think it was great idea using sqlite, it's much faster using that

#839 takes a step towards #833 with not as many performance gains but a far, far simpler solution and much easier maintenance path. Depending on the environment and the default behavior of geopandas to use fiona or pyogrio, the performance improvement may vary. In anycase, the pyogrio library *should* be more performant for the simple case of opening the geopackage to read the list of layers.

AminTorabi-NOAA:
> > Can you also make a PR for #833. I think it was great idea using sqlite, it's much faster using that
> 
> #839 takes a step towards #833 with not as many performance gains but a far, far simpler solution and much easier maintenance path. Depending on the environment and the default behavior of geopandas to use fiona or pyogrio, the performance improvement may vary. In anycase, the pyogrio library _should_ be more performant for the simple case of opening the geopackage to read the list of layers.
> 
> In the meantime, the sqlite3 option is still out there for later refactoring #604 :)

I made a PR with `pyogrio` that significantly improves the speed. However, we're restricted in the packages we can use since they need to be installed via `pip` on WCOSS. I'm following up to see if we can get approval to use that package. The good thing about SQLite, though, is that it comes with Python and doesn't require a pip install

PR #839, opened 2024-09-03
remove fiona dep, require non-dev geopandas

JoshCu:
Spin off from #833 
### Less intrusive change to remove the dependency on fiona
Until wheels for fiona become available on arm for rockylinux, NGIAB can't build with the explicit dependency. 

This change removes the explicit fiona dependency and substitutes it's use with geopandas.
Versions of geopandas prior to 1.0.0 require fiona so I added a version requirement to pyproject.toml. 
The version is there for completeness but not required for NGIAB as it uses the latest version of geopandas without fiona. The version requirement could be dropped to allow pip to install fiona if needed for installations where geopandas<1.0.0 .


JoshCu:
looks like the CI is failing because of the issues fixed here https://github.com/NOAA-OWP/t-route/pull/838

PR #851, opened 2025-01-06
fix: `DataFrame` merge error on `id` field for HF >= 2.2

aaraney:
## Problem:

t-route tries to join two dataframes where one dataframe has two columns with the same name, `id`. A `ValueError` is raised b.c. the column labels are not unique. This seemingly only affects running t-route with HF 2.2.

HF 2.2 introduced a `link` field to the `flowpath-attributes` table. Per the [HF schema docs](https://lynker-spatial.s3-us-west-2.amazonaws.com/hydrofabric/v2.2/hfv2.2-data_model.html), this field is a duplication of the existing `id` field needed for t-route (@shorvath-noaa can you provide background on this?). If t-route detects the `link` field, t-route will rename it from `link` -> `id` causing a column name duplication. Later code joins this dataframe using the `id` field and this causes the aforementioned exception.

## Proposed Solution:

If `link` field is present, drop `id` field _then_ rename `link` to `id`.

This seems to have been addressed in #847 however the change is so minor IMO a smaller PR will be easier to review and resolve issues in using `t-route` with HF 2.2.

### DIFF
--- a/src/troute-network/troute/HYFeaturesNetwork.py
+++ b/src/troute-network/troute/HYFeaturesNetwork.py
@@ -88,8 +88,11 @@ def read_geopkg(file_path, compute_parameters, waterbody_parameters, cpu_pool):
     flowpaths_df = table_dict.get('flowpaths', pd.DataFrame())
     flowpath_attributes_df = table_dict.get('flowpath_attributes', pd.DataFrame())
 
-    # Check if 'link' column exists and rename it to 'id'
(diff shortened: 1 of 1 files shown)


shorvath-noaa:
Yeah, this is one of several updates that t-route needs to be able to run on Hydrofabric v2.2. I don't believe I have the authority anymore to approve/merge PRs on this repo though, maybe @kumdonoaa can?

aaraney:
Okay! Thanks, @shorvath-noaa! I am going to close this for now then.

PR #881, opened 2026-06-16
Apply level-pool routing on virtual flowpaths when a lake feature is present

sclaw:
This PR resolves an oversight in my initial level pool implementation where lake features that did not have an fp_id associated with them would be dropped from routing.  In the updated functionality, any lake features that are associated with a virtual flowpath will lead to a level-pool routing scheme applied to that virtual flowpath.  

This is a notable break from the current conceptual model of NHF T-Route.  Under the current conceptual model, virtual flowpaths that are not associated with a flowpath have flow scaling applied and are not routed on.  To get the new functionality to work, we create temporary div and fp ids for any virtual flowpath that should have level pool routing applied. This promotion allows the vfps to pretend they are flowpaths during network pre-processing.  The vfp links are then removed from the network during waterbody pre-processing.

## Additions

- **_force_headwater_routing** This is where virtual flowpaths are setup to pretend to be flowpaths.  

## Removals

- **_build_div_weighting_matrix** Removed the block where divs where percentages didn't sum to 100 were scaled such that they did sum to 100.  Now, a warning is thrown when percents do not sum to 100.  
Divs that had vfps promoted were having the remaining vfp percentages scaled to 100, and the new div id assigned to the promoted vfp was replacing its percentage with 100.  This led to double counting of lateral flows for the divides where routing was applied to vfps.

## Changes

- Pulled LAKE_ID_FIELD out of the preprocess_waterbodies function to a global variable.
- Specified which fields should be loaded from the lakes layer
- **preprocess_waterbodies** indexed by virtual_fp_id instead of fp_id

## Testing

1. Hot Brook
2. Lake Creek
3. Great Lakes
4. Patuxent
5. (currently waiting on) run_conus_lakes.py


sclaw:
Meant to PR to NGWPC/t-route.

